PR #1, opened 2018-10-02
Update metadata for repo move

raphlinus:
We're moving the repo to the new "fancy-regex" organization. This
patch updates the copyright headers and other metadata.

The contribution guidelines have been updated (adopting the Rust
Code of Conduct), and the README freshened up just a bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ sudo: false
 matrix:
   include:
     - rust: stable
+    - rust: beta
+    - rust: nightly
       addons:
         apt:
           packages:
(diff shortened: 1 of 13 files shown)


raphlinus:
The CI failure seems to be caused by https://github.com/rust-lang/cargo/issues/6015 . I'm wondering whether the best approach for this might be to temporarily disable the code coverage analysis until upstream is fixed, but open to other workarounds.

robinst:
Changes LGTM. Let's see if we can fix the build by running the coverage on nightly: #2

robinst:
I merged the other PR, this should be good to merge now.

On a related topic, what do you think about switching the license from MIT to MIT/Apache-2? That's the combination that all official Rust code has and is fairly popular in the ecosystem too, see: https://rust-lang-nursery.github.io/api-guidelines/necessities.html#c-permissive

PR #2, opened 2018-10-04
travis: Run coverage on nightly instead of stable

robinst:
See https://github.com/rust-lang/cargo/issues/6015

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ sudo: false
 matrix:
   include:
     - rust: stable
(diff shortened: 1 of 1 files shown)


robinst:
Looks like it doesn't work on nightly either: https://travis-ci.com/fancy-regex/fancy-regex/jobs/149630183

But we could merge this anyway, as nightly is allowed to fail.

robinst:
Good idea, I created #3 for that now.

PR #22, opened 2019-09-07
Move deny warnings to Travis build

robinst:
It can be a bit annoying while developing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
 cache: cargo
 
 script:
-  - 'if [[ "$TRAVIS_RUST_VERSION" = 1.32.0 ]]; then RUSTFLAGS="-D warnings" cargo build --verbose; fi'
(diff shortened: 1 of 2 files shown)


robinst:
Hey Raph :)! I got that command from [here](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md#alternatives), not sure either.

Btw planning to release soon, just want to get some more test coverage, especially in `vm`, e.g.: https://codecov.io/gh/fancy-regex/fancy-regex/src/master/src/vm.rs#L322

raphlinus:
I think our source for that may have been https://www.reddit.com/r/rust/comments/8oz7md/make_cargo_fail_on_warning/ . In any case, useful and also surprising there doesn't seem to really be an "official" way to do it. Overriding the RUSTFLAGS environment variable can create problems if the cargo config sets rust flags.

robinst:
Oh thanks, changed it!

PR #33, opened 2019-10-21
Add limit for backtracking in case of catastrophic backtracking

### LINKED ISSUE #6: Add support for backtracking limit
(Moved from previous location: https://github.com/google/fancy-regex/issues/44)

Given that it's possible to run into catastrophic backtracking, I think it would be nice if fancy-regex provided a way to set a limit on backtracking. If the VM reached that limit, it would return an error. Currently, the only way to stop it is to have a separate process and then kill the process.

There's some questions on how that would work:

* Would it be a timeout (e.g. 1 second) or a number that would be incremented and checked when executing instructions
* Would it be controllable via a cargo feature or part of the API when creating a regex (probably introduce a builder). (I'm not sure the feature would work, because that would mean in a crate with multiple dependencies that used fancy-regex, the feature would change the behavior for all those dependent crates.)

@raphlinus's thoughts on this:

> This is a good idea.
>
> I strongly prefer a deterministic approach rather than time. Among other things, running the cloud can introduce random and long delays, which would make tests flaky.
>
> Adding an option at regex creation time sounds good. Note that the API is already designed to fail when MAX_STACK is exceeded, with vm::State::run already returning a Result. So I think it can also be on by default, and won't be too disruptive.
>
> Thanks for agreeing to work on it, and let me know if you want more input.
Interested to hear your thoughts on this @raphlinus, and happy to work on it!

robinst:
Fixes #6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html),
 with the exception that 0.x versions can break between minor versions.
 
+## [unreleased] - 2019-10-19
+### Added
+- Add limit for backtracking so that execution errors instead of running
+  for a long time in case of catastrophic backtracking.
+- Add `RegexBuilder` with `backtrack_limit` to configure the new
+  backtrack limit per regex.
+
 ## [0.2.0] - 2019-10-19
 ### Added
 - More documentation and examples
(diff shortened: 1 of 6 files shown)


robinst:
@raphlinus would appreciate a review. Some questions:

1. Should we have a special value (0?) for the limit to disable it, or do you think it's ok to always have a limit?
2. The current limit of 1 million is arbitrary. The bench I added runs in 81 ms on my (kinda old) laptop. With a limit of 100 thousand, it runs in 8.1 ms (as expected).

PR #36, opened 2019-11-26
Fix backref with multibyte chars

robinst:
The test failed like this before:

    thread 'backref_with_multibyte' panicked at
    'byte index 10 is not a char boundary;
    it is inside '🐱' (bytes 9..13) of `x🐱🐶🐱🐶`',
    src/libcore/str/mod.rs:2068:5


robinst:
@shssoichiro this should fix the problem you have here:

https://github.com/shssoichiro/zxcvbn-rs/blob/v2.0.0/src/matching/mod.rs#L416-L420

PR #38, opened 2019-12-06
Add options for controlling program and dfa size limits

untitaker:
I want to have these options for parsing user-agent regexes in https://github.com/davidarmstronglewis/uap-rs, which parse fine with this crate and onig, but are just too large.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -212,8 +212,8 @@ pub struct SubCaptureMatches<'c, 't> {
 struct RegexOptions {
     pattern: String,
     backtrack_limit: usize,
-    size_limit: Option<usize>,
-    dfa_size_limit: Option<usize>,
+    delegate_size_limit: Option<usize>,
+    delegate_dfa_size_limit: Option<usize>,
 }
 
 impl Default for RegexOptions {
@@ -221,8 +221,8 @@ impl Default for RegexOptions {
         RegexOptions {
             pattern: String::new(),
             backtrack_limit: 1_000_000,
-            size_limit: None,
-            dfa_size_limit: None,
+            delegate_size_limit: None,
+            delegate_dfa_size_limit: None,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -250,6 +254,22 @@ impl RegexBuilder {
         self.0.backtrack_limit = limit;
         self
     }
+
+    /// Set the approximate size limit of the compiled regular expression.
+    ///
+    /// This option is forwarded from the wrapped `regex` crate.
+    pub fn size_limit(&mut self, limit: usize) -> &mut Self {
```
robinst: Can you add some text here that makes it clear that this will only apply to parts of the regex that get delegated, and because of that the actual used size might be size_limit * the number of delegated regexes?

Also, I'm thinking maybe we should add a prefix to this option, like `delegate_size_limit` and `delegate_dfa_size_limit`, what do you think? `size_limit` sounds very generic and maybe at some point we'd want to have a size limit for the VM/non-delegated parts (not sure).

untitaker: The first version of my patch had a `inner_size_limit` etc. Then I thought of the following:

From what I understand `compile_inner` gets [called at least once with a "cooked" regex that is somehow proportional to the size of the original fancy regex](https://github.com/untitaker/fancy-regex/blob/9d650aaf8aaa77e5e1c610ba93bad60c5cd6153d/src/lib.rs#L326).

This would imply to me that, while subsets of the input regex may individually run into this limit too, the `size_limit` applies to the entire regex as well.

lmk if that makes sense

robinst: That's only in the special case where the entire regex can be delegated (if the regex doesn't use *any* fancy features). When there are fancy features like backrefs or look-arounds, only parts of the regex will be delegated. For some regexes, that might mean 3/4 delegates.

untitaker: ok I changed it to `delegate_*`

untitaker:
Thank you very much @robinst for the blazing fast review! Could you issue a new release of fancy-regex as well?

robinst:
Thanks! Released 0.3.1: https://github.com/fancy-regex/fancy-regex/blob/master/CHANGELOG.md#031---2019-12-09

Issue #34, opened 2019-10-30
Named capture groups


robinst:
Named capture groups have been requested as a feature here: https://www.reddit.com/r/rust/comments/djv01f/z/f48r5ao

Should not be too hard to implement:

First, extend parser to parse named groups. Store it in a `HashMap<String, usize>`.

Then add API to look up named groups using the name. Use the `HashMap` to look up the group index first.

Referencing named groups in backrefs doesn't have to be implemented as part of this (but could be).

#### Syntax

Oniguruma uses the syntax `(?<name>subexp)` whereas regex crate uses `(?P<name>subexp)`. We should probably support both.

robinst:
See https://github.com/fancy-regex/fancy-regex/pull/55

robinst:
@rxt1077 @johnw42 released in version 0.4.0 now: https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md#040---2020-09-27

Thank you both! 🎉

PR #47, opened 2020-06-01
Initial named group support

rxt1077:
I implemented _basic_ named groups support based on your comments in issue #34 . Everything is done in parser.rs by keeping keep track of the group number being parsed and using a HashMap to map named group IDs to the numbers.

The result was being able to pass one more internal test and eight more onigurama tests. I'm a little confused as to how (why?) they support some things like non-unique group names. I'm also struggling to grasp the difference between "\g" and "\k" in this context.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -37,6 +37,8 @@ pub enum Error {
     InvalidBackref,
     /// Regex crate error
     InnerError(regex::Error),
+    /// Couldn't find matching '>' in named group
+    InvalidGroupName,
 
     // Run time errors
     /// Max stack size exceeded for backtracking while executing regex.
@@ -81,6 +83,7 @@ impl fmt::Display for Error {
             Error::StackOverflow => write!(f, "Max stack size exceeded for backtracking"),
             Error::BacktrackLimitExceeded => write!(f, "Max limit for backtracking count exceeded"),
             Error::__Nonexhaustive => unreachable!(),
+            Error::InvalidGroupName => write!(f, "Could not parse group name"),
         }
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -41,6 +41,8 @@ pub enum Error {
     InvalidGroupName,
     /// Invalid group id in escape sequence
     InvalidGroupNameBackref(String),
+    /// Once named groups are used you cannot refer to groups by number
+    NamedBackrefOnly,
 
     // Run time errors
     /// Max stack size exceeded for backtracking while executing regex.
@@ -87,6 +89,7 @@ impl fmt::Display for Error {
             Error::__Nonexhaustive => unreachable!(),
             Error::InvalidGroupName => write!(f, "Could not parse group name"),
             Error::InvalidGroupNameBackref(s) => write!(f, "Invalid group name in back reference: {}", s),
+            Error::NamedBackrefOnly => write!(f, "numbered backref/call is not allowed. (use name)"),
         }
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -39,6 +39,8 @@ pub enum Error {
     InnerError(regex::Error),
     /// Couldn't find matching '>' in named group
     InvalidGroupName,
+    /// Invalid group id in escape sequence
+    InvalidGroupNameBackref(String),
 
     // Run time errors
     /// Max stack size exceeded for backtracking while executing regex.
@@ -84,6 +86,7 @@ impl fmt::Display for Error {
             Error::BacktrackLimitExceeded => write!(f, "Max limit for backtracking count exceeded"),
             Error::__Nonexhaustive => unreachable!(),
             Error::InvalidGroupName => write!(f, "Could not parse group name"),
+            Error::InvalidGroupNameBackref(s) => write!(f, "Invalid group name in back reference: {}", s),
         }
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -37,7 +37,7 @@ pub enum Error {
     InvalidBackref,
     /// Regex crate error
     InnerError(regex::Error),
-    /// Couldn't find matching '>' in named group
+    /// Couldn't parse group name
     InvalidGroupName,
     /// Invalid group id in escape sequence
     InvalidGroupNameBackref(String),
--- a/src/error.rs
+++ b/src/error.rs
@@ -89,7 +89,7 @@ impl fmt::Display for Error {
             Error::__Nonexhaustive => unreachable!(),
             Error::InvalidGroupName => write!(f, "Could not parse group name"),
             Error::InvalidGroupNameBackref(s) => write!(f, "Invalid group name in back reference: {}", s),
-            Error::NamedBackrefOnly => write!(f, "numbered backref/call is not allowed. (use name)"),
+            Error::NamedBackrefOnly => write!(f, "Numbered backref/call not allowed because named group was used, use a named backref instead"),
         }
     }
 }
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -24,6 +24,7 @@ use bit_set::BitSet;
 use regex::escape;
 use std::str::FromStr;
 use std::usize;
+use std::collections::HashMap;
 
 use crate::codepoint_len;
 use crate::Error;
@@ -44,6 +45,8 @@ pub(crate) struct Parser<'a> {
     re: &'a str, // source
     backrefs: BitSet,
     flags: u32,
+    named_groups: HashMap<String, usize>,
+    curr_group: usize,  // need to keep track of which group number we're parsing
 }
 
 impl<'a> Parser<'a> {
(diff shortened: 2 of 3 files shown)


keith-hall:
The difference between `\k` and `\g` is quite simple: `\g` matches the text that was previously matched, `\k` applies the regex pattern from the group again (i.e. `\k` is recursion as opposed to a back reference) and so can match something different:
https://rubular.com/r/Jx0hUzZ0fgRozV

rxt1077:
Thanks for the help and the example, but is it the other way around where`\k` is the backref and `\g` is the subcall?

https://github.com/stedolan/jq/wiki/Docs-for-Oniguruma-Regular-Expressions-(RE.txt)

keith-hall:
Yes, apologies - I mixed them up - you are right :)

src/parse.rs
```diff
@@ -258,6 +263,13 @@ impl<'a> Parser<'a> {
                 }
             }
             return Err(Error::InvalidBackref);
+        } else if b == b'k' {
+            if let Some((id, skip)) = parse_id(self.re, ix) {
+                if let Some(group) = self.named_groups.get(id) {
+                    return Ok((ix + skip, Expr::Backref(*group)));
+                }
+            }
+            return Err(Error::InvalidGroupName);
```
robinst, line 272: Let's use a different error for this case, something like `InvalidGroupNameBackref`

src/parse.rs
```diff
@@ -1162,7 +1197,6 @@ mod tests {
 
     #[test]
     fn unknown_flag() {
-        assert_error("(?<name>[-*_])", "Unknown group flag: (?<");
```
robinst, line 1165: Can you add some specific unit tests for named groups in this file too? I'd like to see the error cases covered too (backref before group name defined, etc).

src/parse.rs
```diff
@@ -621,6 +642,20 @@ fn parse_decimal(s: &str, ix: usize) -> Option<(usize, usize)> {
     usize::from_str(&s[ix..end]).ok().map(|val| (end, val))
 }
 
+// finds the an ID between < and > and returns (id, skip) where skip
+// is how far ahead we should jump ix (how much of the string we used)
+fn parse_id(s: &str, ix: usize) -> Option<(&str, usize)> {
+    if let Some(start) = s[ix..].find('<') {
+        if let Some(end) = s[ix..].find('>') {
```
robinst, line 649: Does Oniguruma have any restrictions on what is allowed as the name?

src/parse.rs
rxt1077: I agree. I'll work on this.

src/parse.rs
rxt1077: Certainly.

src/parse.rs
rxt1077: Yes, I believe group names need to start with a letter and then can only be letters and numbers. I'll find out for sure and tighten this up.

rxt1077:
I'd be happy to make the improvements. I'll need to dig a little deeper to learn about the API and it may take me a little while. My work schedule makes the end of the weeks really busy.

Thank you for looking it over and for the feedback. I really appreciate it.

robinst:
Cool. To be clear in case it wasn't: The new API for getting matches via group name doesn't have to be part of this PR, just address the other things and then we can merge this, and do the new API in another PR.

src/parse.rs
rxt1077: Group names have to be made of \w's and/or underscores. I tried to make things work correctly with unicode by using Rust's char_indices function to iterate and is_alphanumeric() to check.

rxt1077:
Understood about the API. In that case I think the last thing I need to nail down is:

> Backreference by number is forbidden if any named group is defined and ONIG_OPTION_CAPTURE_GROUP is not set.

Once I get that done, I'll leave the API stuff to another PR.

rxt1077:
OK, last one done. Let me know if additional changes are needed and thank you for looking at this.

src/parse.rs
```diff
@@ -621,6 +650,41 @@ fn parse_decimal(s: &str, ix: usize) -> Option<(usize, usize)> {
...
+            start = ix + i + 1;
+            break;
+        }
+    }
+    if start == 0 {
+        return None;
+    }
+    
+    // look for the end of the ID while making sure the chars are \w
+    let mut end = 0;
```
robinst, line 670: I think we could write the below in a more concise way:

```
    if let Some(len) = s[start..].find(|c: char| !c.is_alphanumeric() && c != '_') {
        let end = start+len;
        if len > 0 && s[end..].starts_with('>') {
            Some((&s[start..end], end-ix+1))
        } else {
            None
        }
    } else {
        None
    }
```

src/parse.rs
```diff
@@ -251,13 +256,28 @@ impl<'a> Parser<'a> {
         let mut end = ix + 1 + codepoint_len(b);
         let mut size = 1;
         if is_digit(b) {
+            // don't allow numbered groups if named groups have been used
+            if self.named_groups.len() > 0 {
+                return Err(Error::NamedBackrefOnly);
+            }
```
robinst, line 262: I think we need to check this at the end of parsing, otherwise this would not be detected as a syntax error:

`(a)\1(?<name>b)`

src/parse.rs
```diff
@@ -621,6 +650,41 @@ fn parse_decimal(s: &str, ix: usize) -> Option<(usize, usize)> {
...
 }
 
+// finds the an ID between < and > and returns (id, skip) where skip
+// is how far ahead we should jump ix (how much of the string we used)
+// IDs have to be made up of word characters (\w)
+fn parse_id(s: &str, ix: usize) -> Option<(&str, usize)> {
+    // find the start of the ID
+    let mut start = 0;
+    for (i, c) in s[ix..].char_indices() {
+        if c == '<' {
```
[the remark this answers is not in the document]
rxt1077: Good catch. It actually simplifies things to realize that IDs must start with '<'. No searching needed.

src/parse.rs
rxt1077: Very cool. I didn't know about closures as Patterns for find.

src/parse.rs
rxt1077: Thank you for all of the help. I added a test for this and now check for mixed numeric / named backrefs at the end of parsing. Unfortunately it means I had to go through the re again rather than detect the issue as it was being parsed the first time, but I couldn't think of another way without beginning to work outside of parser.rs.

src/parse.rs
```diff
@@ -79,6 +84,14 @@ impl<'a> Parser<'a> {
             }
             return Ok((ix, Expr::Alt(children)));
         }
+        // can't have numeric backrefs and named backrefs
+        for escape in self.re.split('\\').skip(1) { // re can't start with a backref
+            if escape.starts_with(char::is_numeric) {
+                if self.named_groups.len() > 0 {
```
robinst, line 90: Let's put this condition outside. If there's no named groups, there's no need to do any of the checking.

src/parse.rs
```diff
@@ -79,6 +84,14 @@ impl<'a> Parser<'a> {
             }
             return Ok((ix, Expr::Alt(children)));
         }
+        // can't have numeric backrefs and named backrefs
+        for escape in self.re.split('\\').skip(1) { // re can't start with a backref
+            if escape.starts_with(char::is_numeric) {
```
robinst, line 89: I don't think this handles things like `\\1` correctly (that's a literal backslash, followed by a literal 1).

To make this 100% correct (and faster), I would add another field in the parser, something like `numeric_backrefs` that is set to `true` as soon as a single numeric backref is used. Then we can just check that here.

src/error.rs
```diff
@@ -81,6 +87,9 @@ impl fmt::Display for Error {
             Error::StackOverflow => write!(f, "Max stack size exceeded for backtracking"),
             Error::BacktrackLimitExceeded => write!(f, "Max limit for backtracking count exceeded"),
             Error::__Nonexhaustive => unreachable!(),
+            Error::InvalidGroupName => write!(f, "Could not parse group name"),
+            Error::InvalidGroupNameBackref(s) => write!(f, "Invalid group name in back reference: {}", s),
+            Error::NamedBackrefOnly => write!(f, "numbered backref/call is not allowed. (use name)"),
```
robinst, line 92: Style doesn't match other messages, let's change to `Numbered backref/call not allowed because named group was used, use a named backref instead`

(Also makes it clearer.)

robinst:
Ping @rxt1077 just the last bit of changes needed to push this over the finish line

src/parse.rs
rxt1077: Agreed, that makes a lot more sense. I think the original reason I avoided this was that I didn't want to make the parser mutable inside `parse_escape`. It makes things a lot easier though.

rxt1077:
Sorry, this fell completely off my radar. Thank you for the excellent suggestions and your patience.

robinst:
Thanks!

Looks like there's a failing test, but I'm pretty sure it's caused by this fix in regex (which we pull in because we don't have `Cargo.lock` committed): https://github.com/rust-lang/regex/blob/master/CHANGELOG.md#138-2020-05-28

I'll merge and fix afterwards :)

robinst:
@rxt1077 are you going to work on adding the new API to look up capture groups as well? Should be simple after this :).

PR #55, opened 2020-09-08
Named backrefs

johnw42:
I'm adding three semi-independent features in this PR. Let me know if you want me to split it up.

- Access to named capture groups
- Python-compatible syntax for named backrefs
- An method like regex::Captures::expand, with an alternate version using Python-compatible syntax


johnw42:
It looks like the failures being reported are because the CI setup isn't using the latest stable Rust. The remaining Clippy issue is actually caused by a fix I made for a warning from the current version of Clippy.

src/lib.rs
```diff
@@ -718,11 +876,8 @@ impl Expr {
                     buf.push_str("(?:");
                 }
 
-                let is_empty = |e: &Expr| match e {
-                    Expr::Empty => true,
-                    _ => false,
-                };
-                let contains_empty = children.iter().any(&is_empty);
+                let is_empty = |e: &Expr| matches!(e, Expr::Empty);
```
robinst, line 879: This is the reason the build is failing, `matches!` doesn't exist on Rust 1.32.0 which is the oldest version we currently support in fancy-regex. Mind rewriting it to a match?

src/lib.rs
```diff
@@ -452,11 +487,25 @@ impl Regex {
...
+            .capture_names()
+            .enumerate()
+            .filter_map(|(i, name)| name.map(|name| (name.to_string(), i)))
+            .collect();
+        // let mut names = HashMap::new();
+        // for (i, name) in self.capture_names().enumerate() {
+        //     if let Some(name) = name {
+        //         names.insert(name, i);
+        //     }
+        // }
```
robinst, line 500: Old code, delete?

robinst (commented):
This looks excellent, I've done a cursory first review and left some inline comments. Looks like you can fix the build by removing usage of `matches!`.

Also, would you mind splitting off the expand code into its own PR?

I'll do a full review tomorrow.

src/lib.rs
```diff
@@ -155,7 +173,10 @@ const MAX_RECURSION: usize = 64;
 pub struct RegexBuilder(RegexOptions);
 
 /// A compiled regular expression.
-pub struct Regex(RegexImpl);
+pub struct Regex {
+    inner: RegexImpl,
+    group_names: Arc<NamedGroups>,
```
robinst, line 178: Why the `Arc`?

Edit: Ah, I see, the regex crate uses an `Arc` too. But it actually makes use of it, by `clone` ing the arc in the `captures` method: https://docs.rs/regex/1.3.9/src/regex/re_unicode.rs.html#315

Can you:

1. Change `Captures`'s `names` to use an `Arc` too
2. Change your implementation of `captures_from_pos` to use `names: self.group_names.clone(),` instead? That will save a bunch of unnecessary `String` allocations.

src/lib.rs
```diff
@@ -181,7 +202,10 @@ pub struct Match<'t> {
 
 /// A set of capture groups found for a regex.
 #[derive(Debug)]
-pub struct Captures<'t>(CapturesImpl<'t>);
+pub struct Captures<'t> {
+    inner: CapturesImpl<'t>,
+    names: HashMap<String, usize>,
```
robinst, line 207: Looks like you could use `NamedGroups` here too?

tests/captures.rs
```diff
@@ -69,12 +105,14 @@ fn captures_from_pos_looking_left() {
...
 }
 
+#[cfg_attr(feature = "track_caller", track_caller)]
 fn captures<'a>(re: &str, text: &'a str) -> Captures<'a> {
     let regex = common::regex(re);
     let result = regex.captures(text);
     assert_captures(result)
 }
 
+#[cfg_attr(feature = "track_caller", track_caller)]
```
robinst, line 115: I didn't realize `track_caller` could be used for that, great idea! :)

robinst (changes requested):
Ok, some small comments and some changes needed in `Captures` and `captures_from_pos`, otherwise looks good!

src/lib.rs
johnw42: Done. I also change the field name to `named_groups` everywhere to be consistent with the existing code.

robinst:
Perfect, thanks!

PR #56, opened 2020-09-09
Added expansion APIs

johnw42:
Adds a method similar to regex::Captures::expand, an alternate version using Python-style syntax, and a API for flexible expansion string syntax.

Not expected to build with PR #55.

### DIFF
--- a/src/expand.rs
+++ b/src/expand.rs
@@ -1,24 +1,21 @@
 use crate::parse::{parse_decimal, parse_id};
 use crate::Captures;
+use std::borrow::Cow;
 use std::io;
+use std::mem;
 
 /// A set of options for expanding a template string using the contents
 /// of capture groups.  Create using the `builder` method.
 #[derive(Debug)]
-pub struct Expander<'a> {
+pub struct Expander {
     sub_char: char,
-    delimiters: Option<Delimiters<'a>>,
+    open: &'static str,
+    close: &'static str,
     allow_undelimited_name: bool,
     strict: bool,
 }
 
-#[derive(Debug)]
-struct Delimiters<'a> {
-    open: &'a str,
-    close: &'a str,
-}
-
-impl Expander<'static> {
+impl Expander {
     /// Returns an expander that uses Python-compatible syntax.
     ///
     /// Expands all instances of `\num` or `\g<name>` in `replacement`
@@ -26,7 +23,7 @@ impl Expander<'static> {
     /// them to the `dst` buffer given.
     ///
     /// `name` may be an integer corresponding to the index of the
-    /// capture group (counted by order of opening parenthesis where `0` is the
+    /// capture group (counted by order of opening parenthesis where `\0` is the
     /// entire match) or it can be a name (consisting of letters, digits or
     /// underscores) corresponding to a named capture group.
     ///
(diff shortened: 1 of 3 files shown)


src/expand.rs
```diff
@@ -0,0 +1,140 @@
+use crate::parse::{parse_decimal, parse_id};
+use crate::{Captures, Error, Result};
+
+/// A set of options for expanding a template string using the contents
+/// of capture groups.  Create using the `builder` method.
+#[derive(Debug)]
+pub struct Expander<'a> {
```
robinst, line 7: I'm not sure if this should be public API or not. Do you have a use case where you need other delimiters than the ones in `expand` and `expand_backslash`? If not, I'd probably keep this private for now.

robinst (changes requested):
I merged the other PR, you can now rebase this onto main and fix the build.

src/expand.rs
johnw42: I'm not sure either, TBH. I _sort of_ have a use case. Basically I've noticed that while developers mostly treat regex syntax and the associated replacement/expansion syntax as implementation details, they end up being exposed to end users pretty often in things like text editors and command-line tools. My `expand_backslash` method is there because I have a Python program that exposes the syntax as part of its UI, and I wanted to port it to Rust without changing the UI.

I think the basic `expand` method is important to make `fancy_regex` a more complete drop-in replacement for `regex`, and I think either `replace_backslash` should exist, or `Expander` should be public, but probably not both. I like the idea of having an `Expander` type to accommodate multiple syntaxes, but the builder interface is probably overkill, so perhaps it should just have factory methods like `Expander::default()`, `Expander::python()`, etc.

Another nice thing about having an `Expander` type is that it's a place where you can put more expansion methods. The fact that `Captures::expand` has to append to an existing `String` makes it clunky for common use cases. That signature needs to be supported for compatibility with the `regex` create, and I can see why it would sometimes be preferred for performance, but I think most users would prefer a method that returns a new `String`.

I uploaded a new commit that adds the factory methods and changes the expansion methods a bit. If you like that design, I think I'll remove `expand_blackslash` and `ExpanderBuilder`, and update the tests accordingly.

robinst: Thanks for that, that makes sense to me :)!

My initial reaction was that if we have `expand_backslash` we don't need `Expander::python`. But we do, because a user might want to use it with `io::Write`, not `&mut String`. If we get rid of `expand_backslash`, the user can't do `&mut String` without the complicated `Cursor` conversion. But having both still doesn't sit right with me (and also that their naming is not consistent).

How about these (small) changes:

1. On `Captures`, keep only `expand`, for regex compatibility. Add a note to the docs there to see `Expander` for others usages.
2. On `Expander`, add an additional method that can append to a `&mut String`. That then covers all possible append use cases in the "advanced" API.
3. I'm not sure about the naming of the 3 expand methods on `Expander` yet. Some ideas:
   1. `expand -> String`, `expand_to(&mut String)` (or `expand_append(&mut String)`) and `expand_write(impl io::Write)`
   2. `expanded -> String`, `expand(&mut String)` and `expand_to(impl io::Write)`

I think I like the first one. What do you think?

src/lib.rs
```diff
@@ -602,6 +606,62 @@ impl<'t> Captures<'t> {
...
+    /// If `group` isn't a valid capture group (whether the name doesn't exist
+    /// or isn't a valid index), then it is replaced with the empty string.
+    ///
+    /// The longest possible name is used. e.g., `$1a` looks up the capture
+    /// group named `1a` and not the capture group at index `1`. To exert more
+    /// precise control over the name, use braces, e.g., `${1}a`.
+    ///
+    /// To write a literal `$`, use `$$`.    
+    pub fn expand(&self, replacement: &str, dst: &mut String) {
+        let mut cursor = Cursor::new(mem::take(dst).into_bytes());
```
robinst, line 626: To make this work on Rust 1.32:

```suggestion
        let mut cursor = Cursor::new(mem::replace(dst, String::new()).into_bytes());
```

robinst (changes requested):

src/lib.rs
johnw42: For that method names, I don't like mashing two verbs together. How about `expansion`, `append_expansion`, and `write_expansion`? It's kind of unfortunate that the actual "append" method for string is called `extend`; `extend_expansion` doesn't make sense and something like `extend_with_expansion` just seems way too long.

johnw42: I uploaded a new patch with the names I proposed, and added `Expander::quote`, since that's another feature I'm familiar with from Python.

The `set_strict` method could be eliminated if you don't think reporting errors in the expansion template is worthwhile. In that case, `expansion` could return `String` and `append_expansion` could return `()`.

src/expand.rs
```diff
@@ -0,0 +1,175 @@
+use crate::parse::{parse_decimal, parse_id};
+use crate::Captures;
+use std::borrow::Cow;
+use std::io;
+use std::mem;
+
+/// A set of options for expanding a template string using the contents
+/// of capture groups.  Create using the `builder` method.
```
robinst, line 8: I didn't mind having the builder btw, you can add it back if you want. Otherwise, remove this from the docs :).

src/expand.rs
```diff
@@ -0,0 +1,175 @@
...
+        Ok(String::from_utf8(cursor.into_inner()).expect("expansion is UTF-8"))
+    }
+
+    /// Appends the expansion produced by `expansion` to `dst`.  Possibly more efficient
+    /// than calling `expansion` directly.
+    pub fn append_expansion<'t>(
+        &self,
+        dst: &mut String,
+        captures: &Captures<'t>,
+        template: &str,
```
robinst, line 94: I like the new names :)!

One more comment on the API: Should we put the `template` arguments before `captures`?

Why? Because that would be similar to how `write!` is used, e.g.: `write!(f, "Regex error: {}", e)`. So, destination first, then template string, then arguments (which in our case come from the captures). What do you think?

src/expand.rs
johnw42: I decided the builder API did exposes too many implementation details without adding useful generality. In particular, I think Perl and Sed syntax might be good additions at some point, and I don't thing either could be reproduced faithfully with the builder API. The default syntax is perfectly fine for any use case that doesn't require compatibility with another implementation, and adding an API that _almost_ allows emulating another implementation seems counterproductive.

src/expand.rs
johnw42: Seems reasonable to me.

robinst:
Thanks! Are you planning to work on anything else :)? Just curious (and there's more work to be done, always).

robinst:
(Waiting for https://github.com/fancy-regex/fancy-regex/pull/57 to get in before cutting a release.)

johnw42:
Maybe. I've been trying to come up with an API that allows detecting errors in an expansion template without first needing a regex match. My motivating example is the same script I mentioned before, which is basically just a grep-like tool extended with the ability to do replacements. I want to be able to validate that the replacement text is valid before looking for matches, or in cases where there are no matches.

Actually detecting the errors is easy; that's already done, as you can see by the `expander_errors` test in `captures.rs`. The part I'm having difficulty with is reporting the errors in a useful way that's not overly complicated. The traditional approach is to never report any errors, and just treat any broken expansion syntax as a literal string. I think that should be the default, to conform to developers' expectations.

A maximal approach would be to detect and classify all possible errors in the template and report them all, along with corresponding locations in the input string. Going down that path quickly gets ridiculous, though; the logical conclusion would be to have a method that parses the template in the context of a `Regex` and returns `Result` where the `Err` type is a `Vec` of objects that can be printed as errors like you'd get from a compiler. With this approach, `Expander` becomes an object built from a template string, and the `quote` method has to move somewhere else, because if you need it, you need it before you've even built the template string. That, in turn, means you probably want a separate `ExpanderSyntax` type. Crazy, right?

I'm trying to think of a good compromise between those extremes. Here's where I'm at right now:

- Leave the basic API mostly unchanged, but get rid of the `set_strict` method and remove `io::Result` from the return types of `expansion` and `append_expansion`, because error checking will be a separate step. (It's still needed for `write_expansion` to propagate errors from the `io::Write` calls, though.)
- Add a `check` method that takes a `&Regex` and returns a `fancy_regex::Result<()>` value. Errors are reported the same way they're currently reported from `expansion`.

WDYT?

PR #57, opened 2020-09-21
Added: `Match:range()` method

lebensterben:
- These are ported from `regex` crate.
- Added `Match::range()`
- Added description for `Match::new()`
- Implemented `From<Match>` for string slice and `Range<usize>`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -566,6 +566,7 @@ impl<'t> Match<'t> {
         &self.text[self.start..self.end]
     }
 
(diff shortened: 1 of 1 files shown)


robinst:
Hey! Thanks, these look nice. I know the implementations are trivial, but mind adding some tests (maybe in `finding.rs`)?

robinst:
Done in https://github.com/fancy-regex/fancy-regex/commit/ea6366527dad0415b57356c952cb3d8f44f60c2c

PR #58, opened 2020-09-23
Added error checking and fixed append_expansion.

johnw42:
This adds the error checking method I mentioned before, and it fixes a bug in `append_expansion` that causes the initial contents of `dst` to be overwritten.


robinst:
Thanks!

PR #61, opened 2020-10-27
Added escape() function and renamed Expander::quote to Expander::escape.

johnw42:


src/expand.rs
```diff
@@ -75,6 +75,12 @@ impl Expander {
         }
     }
 
+    #[doc(hidden)]
+    #[deprecated(since = "0.3.6", note = "Use `escape` instead.")]
```
robinst, line 79: The last version is 0.4.0, 0.3.6 doesn't exist:

```suggestion
    #[deprecated(since = "0.4.0", note = "Use `escape` instead.")]
```

robinst (changes requested):
One minor nit, otherwise looks good!

florianpircher (approved):
Looks good.

PR #67, opened 2021-02-13
add find_iter, captures_iter (and find_from_pos)

### LINKED ISSUE #59: Add Regex::captures_iter
See https://docs.rs/regex/1.3.9/regex/struct.Regex.html#method.captures_iter

Currently, you will have to call `captures_from_pos` repeatedly to get this. Would be good to add this as API. Other things such as the replace API would also make use of it (#49).

bminixhofer:
Hi!

This PR would fix #59 by adding `find_iter` and `captures_iter`. I also had to add `find_from_pos` which is called repeatedly by `find_iter` now. The implementations are adapted from the regex crate with minimal modifications.

Some things worth noting:
- There is some code duplication between the `Iterator` impl for `Matches` and `CaptureMatches`. This is also the case in the regex crate and I didn't see an easy way to avoid this.
- I had to copy the `next_utf8` method from the regex crate to move forward in case of an empty match. I made a note in the doc string.

Before this is merged, the current status in the Readme should also be updated.

Please let me know what you think!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -203,6 +203,123 @@ pub struct Match<'t> {
     end: usize,
 }
 
+/// Returns the smallest possible index of the next valid UTF-8 sequence
+/// starting after `i`.
+/// Adapted from a function with the same name in the `regex` crate.
+fn next_utf8(text: &str, i: usize) -> usize {
+    let b = match text.as_bytes().get(i) {
+        None => return i + 1,
+        Some(&b) => b,
+    };
+    let inc = if b <= 0x7F {
+        1
+    } else if b <= 0b110_11111 {
+        2
+    } else if b <= 0b1110_1111 {
+        3
+    } else {
+        4
+    };
+    i + inc
+}
+
+/// An iterator over all non-overlapping matches for a particular string.
+///
+/// The iterator yields a `Match` value. The iterator stops when no more
+/// matches can be found.
+///
+/// `'r` is the lifetime of the compiled regular expression and `'t` is the
+/// lifetime of the matched string.
+#[derive(Debug)]
+pub struct Matches<'r, 't> {
+    re: &'r Regex,
+    text: &'t str,
+    last_end: usize,
+    last_match: Option<usize>,
+}
+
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -203,6 +203,145 @@ pub struct Match<'t> {
...
+    let inc = if b <= 0x7F {
+        1
+    } else if b <= 0b110_11111 {
+        2
+    } else if b <= 0b1110_1111 {
+        3
+    } else {
+        4
+    };
+    i + inc
```
robinst, line 223: We already have a function called `codepoint_len` which we should be able to use here, to not duplicate this logic.

src/lib.rs
```diff
@@ -418,17 +585,68 @@ impl Regex {
...
+    /// let re = Regex::new(r"(?m:^)(\d+)").unwrap();
+    /// let text = "1 test 123\n2 foo";
+    /// let mat = re.find_from_pos(text, 7).unwrap().unwrap();
+    ///
+    /// assert_eq!(mat.start(), 11);
+    /// assert_eq!(mat.end(), 12);
+    /// ```
+    ///
+    /// Note that in some cases this is not the same as using the `captures`
+    /// methods and passing a slice of the string, see [Regex::captures_from_pos()] for details.
```
robinst, line 608:
```suggestion
    /// Note that in some cases this is not the same as using the `find`
    /// method and passing a slice of the string, see [Regex::captures_from_pos()] for details.
```

robinst (changes requested):
Hey! Thanks for working on this, this looks excellent. There's only some trivial things that should be addressed, after that we can merge and release this.

src/lib.rs
bminixhofer: I fixed this and moved `next_utf8` next to the other utility methods. Also added a test for zero length matches before longer codepoints (s. t. `next_utf8` is expected to increment by more than one).

bminixhofer:
Thanks for the quick review! I addressed those issues and went ahead and updated the status in the Readme.

Issue #72, opened 2021-05-15
Implement Display, Index for feature parity with regex


duckontheweb:
While substituting this crate for `regex` in Stranger6667/jsonschema-rs#220 it came up that there are a couple traits that are implemented on `regex` and not here. These seems like pretty low-hanging fruit and would make it a bit easier to swap this in place of `regex`. These are the ones that I saw, but there may be others:

* [`impl Display for Regex`](https://docs.rs/regex/1.5.4/regex/struct.Regex.html#impl-Display)
* [`impl Index for Captures`](https://docs.rs/regex/1.5.4/regex/struct.Captures.html#impl-Index%3C%26%27i%20str%3E)

I can put in a PR for these 2. I'll take a look at `Clone` and `FromStr` (from #71) as well to see what the level of effort would be there.

robinst:
Released in 0.6.0: https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md#060---2021-05-17

Issue #71, opened 2021-05-14
doesn't impl `clone` and `FromStr`


steveeJ:
These two trait impls would be convenient for building CLI applications that use structs for passing on arguments to subcommands.

Without these I'm using the following workarounds.

Custom parsing function with `StructOpt`. This assumes that `,` isn't used in the pattern itself, which of course is arbitrary and doesn't work in the general case.
```rust
    #[derive(StructOpt, Debug)]
    pub(crate) struct Check {
        #[structopt(long, default_value = "", parse(try_from_str = parse_regexmatch))]
        pub(crate) packages_filters: fancy_regex::Regex,
    }

    fn parse_regexmatch(input: &str) -> Fallible<fancy_regex::Regex> {
        let set = fancy_regex::Regex::new(
            &input
                .split(",")
                .filter(|s| !s.is_empty())
                .collect::<String>(),
        )?;
        Ok(set)
    }
```

Using `Regex::new` with `Regex::as_str()` as input. Being able to just call `Regex::clone` would be nice.
```rust
...
                packages_filters: fancy_regex::Regex::new(self.packages_filters.as_str())
                    .expect("regex is valid"),
...
```

robinst:
Hey! I just released 0.6.0 which should fix your issue: https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md#060---2021-05-17

Let me know if it doesn't.

steveeJ:
Works great, thank you @robinst!

PR #73, opened 2021-05-16
Implement some common traits on Regex & Captures

duckontheweb:
Adds the following implementations on the `Regex` and `Captures` structs. These implementations follow those in the `regex` crate, and in some cases are direct copies.

* `impl fmt::Display for Regex`
* `impl<'t> Index<usize> for Captures<'t>` (index capture groups by `usize`)
* `impl<'t, 'i> Index<&'i str> for Captures<'t>` (index capture groups by name)
* `impl Clone for Regex` (also implements for `RegexImpl`, `vm::Insn`, `vm::Prog`, which `Regex` relies on)

Addresses #72 and at least part of #71 (not sure if it totally covers the need described there).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -424,6 +424,13 @@ impl fmt::Debug for Regex {
     }
 }
 
+impl fmt::Display for Regex {
+    /// Shows the original regular expression
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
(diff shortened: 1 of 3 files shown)


robinst:
Excellent, thanks!

robinst:
Released in 0.6.0: https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md#060---2021-05-17

PR #75, opened 2021-07-01
Add replace*() methods similar to regex crate

### LINKED ISSUE #49: Add Regex::replace (and replacen, replaceall)
I thought this crate would have a replace or replace_all like the rust regex crate does.
Is there any way to do replace with this crate?

intgr:
Closes #49

Most of this code is verbatim copy from the `regex` crate.

License: MIT OR Apache-2.0
Author: Andrew Gallant <jamslam@gmail.com>
Author: Alex Crichton <alex@alexcrichton.com>
Author: Lapinot <peio.borthelle@ens-lyon.fr>
Author: tom <nooberfsh@gmail.com>
Author: Matt Brubeck <mbrubeck@limpet.net>
Author: Marti Raudsepp <marti@juffo.org>

Methods:
* Regex::replace() -- single replacement
* Regex::replace_all() -- replace all non-overlapping matches
* Regex::replacen() -- configurable number of replacements

Traits:
* Replacer -- describes logic for replacing matches

  Trait implementations included for `&str`, `String`, `NoExpand`, `FnMut(&Captures)->str`

Structs:
* NoExpand -- Replacer that performs no substitutions
* ReplacerRef -- by-reference adaptor

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -157,6 +157,7 @@ mod compile;
 mod error;
 mod expand;
 mod parse;
+mod replacer;
 mod vm;
 
 use crate::analyze::analyze;
@@ -166,6 +167,7 @@ use crate::vm::Prog;
 
 pub use crate::error::{Error, Result};
 pub use crate::expand::Expander;
+pub use crate::replacer::{NoExpand, Replacer, ReplacerRef};
 use std::borrow::Cow;
 
 const MAX_RECURSION: usize = 64;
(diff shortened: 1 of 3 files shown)


intgr:
Unfortunately, I had to duplicate `Replacer` trait-related code, instead of re-exporting from `regex`. Because it all depends on the `Captures` struct, which is distinct in `fancy-regex`. Although maybe there is some opportunity to share code that I missed?

src/lib.rs
```diff
@@ -759,6 +761,188 @@ impl Regex {
...
+    /// To write a literal `$` use `$$`.
+    ///
+    /// # Examples
+    ///
+    /// Note that this function is polymorphic with respect to the replacement.
+    /// In typical usage, this can just be a normal string:
+    ///
+    /// ```rust
+    /// # use fancy_regex::Regex;
+    /// # fn main() {
```
robinst, line 798: Just curious is the `fn main` block here necessary? If not, we can remove it to make it a bit cleaner to read.

tests/replace.rs
```diff
@@ -0,0 +1,10 @@
+mod common;
+
+#[test]
+fn replace() {
+    let regex = common::regex(
+        r"\b([sS])uc(?:cs|s?)e(ed(?:ed|ing|s?)|ss(?:es|ful(?:ly)?|i(?:ons?|ve(?:ly)?)|ors?)?)\b",
+    );
+    let result = regex.replace("a sucessful b", "${1}ucce$2");
+    assert_eq!(result, "a successful b");
```
robinst, line 9: Can you add some more tests? It looks like there's none for `replace_all` (not even doctests). Also not sure if all the replacers are covered, it would be good to have coverage of all of them.

robinst (changes requested):
Hey! Thanks, this looks really good already. Just some minor things to address.

intgr:
Thanks, will submit fixes in the evening if I don't forget :)

robinst:
Thanks! Merged, will prepare a release with this.

robinst:
Released in 0.7.0, thanks @intgr: https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md#070---2021-07-12

intgr:
Thanks :)

PR #78, opened 2021-07-26
fix: Panic on incomplete escape sequences in input regexes

### LINKED ISSUE #76: Panic on not common input
Hi! Thank you for this library!

While fuzzing the [jsonschema](https://github.com/Stranger6667/jsonschema-rs/issues/253) crate, I found that `fancy-regex` panics on `\\u` (and a few more similar patterns like `\\U` and `\\x`), with:

```
thread 'main' panicked at 'index out of bounds: the len is 2 but the index is 2', /home/stranger6667/.cargo/registry/src/github.com-1ecc6299db9ec823/fancy-regex-0.6.0/src/parse.rs:394:17
```

**Code to reproduce**:

```rust
fn main() {
    fancy_regex::Regex::new("\\u");
}
```

Though it is an incomplete escape sequence, it will be nice to return `Err` in such cases as in `regex`:

```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    \u
      ^
error: incomplete escape sequence, reached end of pattern prematurely
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Probably, the solution will be to check the bounds before going to `bytes[ix]` inside `Parser::parse_hex`.
Let me know what do you think, I'll be happy to submit a PR to fix this :)

Stranger6667:
Fixes #76

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html),
 with the exception that 0.x versions can break between minor versions.
 
+### Fixed
+
+- Panic on incomplete escape sequences in input regexes
+
 ## [0.7.0] - 2021-07-12
 ### Added
 - `Regex` now has replace methods like the regex crate:
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -390,6 +390,10 @@ impl<'a> Parser<'a> {
 
     // ix points after '\x', eg to 'A0' or '{12345}', or after `\u` or `\U`
     fn parse_hex(&self, ix: usize, digits: usize) -> Result<(usize, Expr)> {
+        if ix >= self.re.len() {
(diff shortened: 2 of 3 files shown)


robinst:
Hey, thanks :). Looks like the build is failing with Rust 1.41.1 because a newer criterion version started depending on something that's not available in 1.41.1.

We need to either:

* Pin the exact version in `Cargo.toml` for criterion
* Commit a `Cargo.lock` with known-working dependencies

I've been doing the former to fix these kinds of problems, but starting to lean towards the latter to make builds more stable.
I'll raise a PR to fix it.

robinst:
Fixed in https://github.com/fancy-regex/fancy-regex/commit/f99e5f6ecad529f994fe61c00ce6a4884f402b42

Stranger6667:
@robinst 

Awesome! Thank you very much :)

PR #79, opened 2021-07-27
Disallow quantifers on lookarounds and other zero-width assertion expressions

### LINKED ISSUE #68: Look-ahead/behind with repeat operator should not compile
From https://github.com/fancy-regex/fancy-regex/issues/66#issuecomment-779820378

The following types of regexes should not compile (but they currently do):

```
(?=a)*
(?!)?
(?=a){1,2}
(?<=a)*
(?<!a)*
```

Note that Oniguruma fails with this error: `target of repeat operator is invalid`

keith-hall:
Fixes #68 and matches Oniguruma behavior

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -149,7 +149,7 @@ impl<'a> Parser<'a> {
                 }
                 _ => return Ok((ix, child)),
             };
-            if let Expr::LookAround(_, _) = child {
+            if !self.is_repeatable(&child) {
                 return Err(Error::TargetNotRepeatable);
             }
             ix += 1;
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -172,6 +175,18 @@ impl<'a> Parser<'a> {
...
 
+    fn is_repeatable(&self, child: &Expr) -> bool {
+        match child {
+            Expr::LookAround(_, _) => false,
+            Expr::Empty => false,
+            Expr::StartText => false,
+            Expr::EndText => false,
+            Expr::StartLine => false,
+            Expr::EndLine => false,
+            _ => true
```
robinst, line 186: Mind running `cargo fmt`? Looks like the build is failing because of a missing comma here.

robinst (changes requested):
Thanks! This looks good. Can you just rebase your changes on master and fix the formatting issue? Then we should have a clean build.

src/parse.rs
keith-hall: done :) maybe worth mentioning this in the contributing guide to potentially save time in future for new contributors? :)

robinst: Yep done :)

PR #81, opened 2021-08-19
Draft: Enhancement on Match and Regex

skanfd:
I found it's hard to inter-operate fancy-regex with my code, and I've tried to invent some new structs and traits to solve the problem but failed (ugly and doesn't work well), that's why I hope to enhance `Match` and `Regex` itself (and I prefer to solve general problems rather than my problems ).

To be specific, I hope to enhance from 4 aspects:

1. two methods on `Regex`

   - `find_iter_from_pos`
   - `captures_iter_from_pos`

2. one method on `Match`

   - ~~`cursor` to find a human-readable cursor position `(row, column)` which counts from 1 not 0. This method tell us where the `Match` matched exactly, similar to `range` but in a more friendly way.~~
   - `text()` to return the text being searched

3. impl<'t> Debug for Match<'t>

   - debug `Match` friendly

4. a set of methods similar to `Regex`'s methods but on `Match` to do "inner search" (I call this to do a further search on the `Match`'s `.text` from `.start`  byte to `.end` byte)

   - `find(&self, re: &'t Regex) -> Result<Option<Match<'t>>> `
   - `find_from_pos(&self, re: &'t Regex, pos: usize) -> Result<Option<Match<'t>>>`
   - `find_iter(&self, re: &'t Regex) -> Matches<'t, 't>`
   - `find_iter_from_pos(&self, re: &'t Regex, pos: usize) -> Matches<'t, 't>`
   - `captures(&self, re: &'t Regex) -> Result<Option<Captures<'t>>>`
   - `captures_from_pos(&self, re: &'t Regex, pos: usize) -> Result<Option<Captures<'t>>>`
   - `captures_iter(&self, re: &'t Regex) -> CaptureMatches<'t, 't>`
   - `captures_iter_from_pos(&self, re: &'t Regex, pos: usize) -> CaptureMatches<'t, 't>`

   

As the author of `fancy-regex`, do you think this enhancement is proper?

I want your advice, thank you!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -201,11 +201,20 @@ enum RegexImpl {
 }
 
 /// A single match of a regex or group in an input text
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq)]
 pub struct Match<'t> {
     text: &'t str,
     start: usize,
     end: usize,
+}
+
+impl Debug for Match<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Match")
+            .field("as_str", &self.as_str())
+            .field("range", &self.range())
+            .finish()
+    }
 }
 
 /// An iterator over all non-overlapping matches for a particular string.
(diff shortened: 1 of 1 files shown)


robinst:
Hey!

I saw this in my inbox but didn't get to review until now. To your specific enhancements:

1. That might be useful in some cases, but I'd like to keep the API as close to the regex crate as possible and not add too many extra methods. The functionality can already be done using `_from_pos` methods.

2. Same as 1.

3. I actually think this might be reasonable, wonder why the regex crate doesn't do it.

4. I think this adds a bit too much additional API for what it does.

skanfd:
I have been thinking about what I need these days, but I found that the changes I made were not appropriate (and for this crate), which is why I closed this pull request.

Anyway, thank you for your reply, and how nice of you to write fancy-regex :)

PR #83, opened 2021-11-28
Update fuzzing harness, use correct name to parse expr

5225225:
I just deleted the fuzzer and regenerated it so that I can use `data: &str`


robinst:
Thanks! We should probably also at least build the fuzzing part as part of GitHub actions builds.

PR #85, opened 2022-02-15
Enable selection of regex crate features

mbrubeck:
Allow users to disable any of the `unicode` and `perf-*` features of the `regex` crate.  Disabling these features can reduce compile time and/or binary size for use cases where these features are not needed.  (All features remain enabled by default.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,11 +11,22 @@ documentation = "https://docs.rs/fancy-regex"
 categories = ["text-processing"]
 
 [features]
+default = ["unicode", "perf"]
 # Enable #[track_caller] in unit tests.
 track_caller = []
+perf = ["regex/perf"]
+perf-dfa = ["regex/perf-dfa"]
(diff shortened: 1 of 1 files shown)


robinst:
Thanks! Out of interest, what are you using fancy-regex for? :)

mbrubeck:
We're looking to use it at work in our mobile (Android/iOS) apps, which are partly implemented in Rust.  We need lookahead for compatibility with existing regexes that are currently processed by a JavaScript app.

edmorley:
@robinst I don't suppose you know when you might publish a new release containing this change?


p.s. Small world @mbrubeck 👋 😄

robinst:
Oh thanks for reminding me :). Version 0.8.0 released with this now 🎉: https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md#080---2022-02-22

edmorley:
@robinst Thank you :-)

PR #88, opened 2022-04-13
Add support for \K keep out

### LINKED ISSUE #87: Support for Keep out \K
There is a method for this, without using lookbehind?

keith-hall:
Fixes #87 

Note: I haven't looked into how oniguruma handles `\K` inside lookarounds yet.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -439,6 +439,9 @@ pub(crate) fn run(
                     if option_flags & OPTION_TRACE != 0 {
                         println!("saves: {:?}", state.saves);
                     }
+                    if state.saves[0] > state.saves[1] {
+                        state.save(0, state.saves[1]);
+                    }
                     return Ok(Some(state.saves));
                 }
                 Insn::Any => {
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -439,8 +439,9 @@ pub(crate) fn run(
                     if option_flags & OPTION_TRACE != 0 {
                         println!("saves: {:?}", state.saves);
                     }
-                    if state.saves[0] > state.saves[1] {
-                        state.save(0, state.saves[1]);
+                    let slot1 = state.get(1);
+                    if state.get(0) > slot1 {
+                        state.save(0, slot1);
                     }
                     return Ok(Some(state.saves));
                 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html),
 with the exception that 0.x versions can break between minor versions.
 
+## [Unreleased]
+### Added
+- Added support for `\K` keep out
+
 ## [0.8.0] - 2022-02-22
 ### Added
 - Allow users to disable any of the `unicode` and `perf-*` features of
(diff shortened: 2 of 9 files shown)


robinst:
> Note: I haven't looked into how oniguruma handles \K inside lookarounds yet.

Just checked, it allows it and it affects the positions:

```
(?<=a\Kb)c matches abc: Some((1, 3))
(?<!a\Kb)c matches axc: Some((2, 3))
a(?=b\Kc) matches abc: Some((1, 1))
a(?!b\Kc) matches abx: Some((0, 1))
```

Here's the commit where oniguruma added support for keep: https://github.com/kkos/oniguruma/commit/8b6c6304f587af8cce1bbc101211ec0531769965

robinst:
Here's those examples as test cases:

```
    assert_eq!(find(r"(?<=a\Kb)c", "abc"), Some((1, 3)));
    assert_eq!(find(r"(?<!a\Kb)c", "axc"), Some((2, 3)));
    assert_eq!(find(r"a(?=b\Kc)", "abc"), Some((1, 1)));
    assert_eq!(find(r"a(?!b\Kc)", "abx"), Some((0, 1)));
```

After fixing the const problem by adding `const_size = true;` in analyze, the third case fails. It returns `Some((2, 1))` with this PR. I think the oniguruma result is a bit weird, but seems like all it does is cap the start index to be <= end index (see [code](https://github.com/kkos/oniguruma/blob/879a4e2943bc18a8c165ed9a9d888bf9d8c68d82/src/regexec.c#L3078)).

keith-hall:
Thanks for taking a look! I've updated the branch/PR with your additional tests and `const_size` fix. I think the problem we have with that third case is that we are doing a `Save(0)` inside the lookahead, and then restoring back to `0` after we leave the lookahead.

```dot
digraph G {
  0 [label="0: Lit(\"a\")"];
  0 -> 1;
  1 [label="1: Save(0)"];
  1 -> 2;
  2 [label="2: Lit(\"b\")"];
  2 -> 3;
  3 [label="3: Save(0)"];
  3 -> 4;
  4 [label="4: Lit(\"c\")"];
  4 -> 5;
  5 [label="5: Restore(0)"];
  5 -> 6;
  6 [label="6: End"];
}
```

What are the ways we could solve this?

One idea I was thinking about would be at compile time, when visiting a lookahead, to traverse it first and look for `KeepOut`, and if encountered, emit the `Save(0)` instruction before entering the lookahead.

Alternatively, maybe it would work to introduce another `Insn` which would save but be ignored for `Restore` operations? I haven't thought that one through much yet, admittedly.

keith-hall:
Nevermind, I think I missed the point there :sweat_smile: I have now pushed a commit which caps the start index to be <= end index.

src/vm.rs
```diff
@@ -439,6 +439,9 @@ pub(crate) fn run(
                     if option_flags & OPTION_TRACE != 0 {
                         println!("saves: {:?}", state.saves);
                     }
+                    if state.saves[0] > state.saves[1] {
+                        state.save(0, state.saves[1]);
+                    }
```
keith-hall, line 444: not sure if it's okay to access `saves` by index here, or if we should use `get` instead?
```suggestion
                    let slot1 = state.get(1);
                    if state.get(0) > slot1 {
                        state.save(0, slot1);
                    }
```

robinst: Yeah let's use `get` like in other places.

robinst (commented):
Let's apply your suggested change, other than that, looks good now! Ah, feel free to add an entry to `CHANGELOG.md` as well.

robinst:
Released as 0.9.0, thanks a lot @keith-hall!

PR #95, opened 2022-04-27
Extend docs/comments for \K and \G

robinst:
@keith-hall


keith-hall (approved):
Nice, very clear explanations and a great idea to add links to docs!

PR #97, opened 2022-05-04
Improve parse errors by showing the position they occurred at

keith-hall:
Improve parse errors by showing the position they occurred at. The motivation for this is it becomes easier to debug when adding support for new syntax.


src/analyze.rs
```diff
@@ -167,13 +167,13 @@ impl<'a> Analyzer<'a> {
             }
             Expr::Backref(group) => {
                 if group >= self.group_ix {
-                    return Err(Error::InvalidBackref);
+                    return Err(Error::InvalidBackref(0));
```
keith-hall, line 170: I wasn't sure what position to return when we return an error during analyze or compile time which can also occur at parse time.

src/error.rs
```diff
@@ -68,31 +70,31 @@ impl fmt::Display for Error {
         // We should make these more helpful, e.g. by including the parts of the regex that lead to
```
keith-hall, line 70: We may be able to remove this comment now :)

src/analyze.rs
keith-hall: fixed by adding separate error types for parse time errors, analyze/compile time errors and runtime errors

src/error.rs
keith-hall: comment removed :)

keith-hall:
> ---- oniguruma stdout ----
thread 'oniguruma' panicked at 'Expected failure differed for test, change it in the ignore file: x2("(?:(?'name'a)|(?'name'b))(?('name')c|d)e", "ace", 0, 3);
Expected: Compile failed: ParseError(5, UnknownFlag("(?'"))
Actual  : Compile failed: ParseError(5, UnknownFlag("(?\'"))
', tests/oniguruma.rs:223:13

hmm, if I change the comment to include the backslash, then the test fails locally... Not sure why CI is behaving differently here...

robinst:
Looks like it's only on 1.42.0, hmm. I wonder if formatting of errors changed in a later version? Have you tried running the tests with 1.42.0?:

```
$ rustup install 1.42
$ cargo +1.42 test
```

robinst:
This change looks useful! Note that the regex crates has more detailed errors with spans (start and end positions) and pretty printing (see an example for a single-character error [here](https://github.com/rust-lang/regex/blob/a2a393f1ff9ff631d4be818de7488e26b007ae63/regex-syntax/src/error.rs#L317-L319)). I don't think we should put in the effort to implement that just yet, just wanted to point it out.

keith-hall:
Good call. Yup, I can reproduce the test failure locally with rust 1.42.

keith-hall:
Looks like this was fixed in Rust 1.53: https://github.com/rust-lang/rust/issues/83046
This is affecting us because we use a debug format string at https://github.com/fancy-regex/fancy-regex/blob/484897599504c16a236949de1d321180efe5d882/tests/oniguruma.rs#L156

I'm pretty sure we don't want to bump our MSRV so much just for this... I wonder what the best way forward is, any ideas?

robinst:
Yeah.. I think we can just do something like this, with a comment about removing it when we bump from 1.42:

```diff
-            assert!(failure.starts_with(expected_failure),
+            assert!(failure.starts_with(expected_failure) || failure.replace("\\", "").starts_with(expected_failure),
```

keith-hall:
Nice, that works, thanks - done :) We could also rely on the [rustversion crate](https://crates.io/crates/rustversion) for it to be self-documenting, but probably not worth adding another dependency just for that ;)

robinst:
Thanks @keith-hall! Not gonna release this just yet, unless you want me to.

keith-hall:
No, it's not necessary, thanks :) I will work on the next feature :)

PR #99, opened 2022-07-15
Add support for conditionals

keith-hall:
This PR adds support for Oniguruma's "backreference validity checker" functionality, plus conditionals - basically if/then/else expressions.

To achieve this, I ~~only needed to add~~ added one additional `Expr` variant for the parser and a corresponding `Insn` variant for the VM.
I have called them `BackrefExistsCondition`, which takes the capture group number as a parameter. (Named capture groups are automatically converted to numbered ones in the parser. Meanwhile, I noticed there was an `Expr::NamedBackref` which was never used, so I took the opportunity to remove it.)

This instruction is marked as hard, and const size - always zero. I added some comments in the parser to hopefully explain how it works, but happy to clarify anything here if needed. :)

----
EDIT to add:

After some investigation, I realized I also needed to add a `Expr` for the conditional expression.


src/parse.rs
```diff
@@ -679,6 +681,78 @@ impl<'a> Parser<'a> {
...
+        }
+        let inner_condition = if let Expr::Backref(group) = condition {
+            Expr::BackrefExistsCondition(group)
+        } else {
+            condition
+        };
+
+        Ok((
+            end + 1,
+            Expr::Alt(vec![Expr::Concat(vec![inner_condition, if_true]), if_false]),
```
keith-hall, line 749: Unfortunately there is a bug with this approach - the idea of being able to use `Expr::Alt` for the inner condition only works if the inner condition is the `BackrefExistsCondition`. This is why the test `assert_no_match(r"^(?(\d)abc|\d!)$", "5!");` fails. `\d` from the condition matches, but `abc` doesn't, so it should fail. But due to the alternation, it tries `\d!`, which passes. Instead, the `\d` from the condition should have consumed the `5` already. I will try to think how best to solve this, maybe with Split somehow?

keith-hall: Okay, so I worked out that the example in my previous comment wasn't the best, as only lookarounds or backref validity checks can be used in the conditional expression. I have updated the code and tests accordingly. But I still need to fix it to ensure we don't try the false branch when the truth branch fails after the condition succeeded.

keith-hall: > Okay, so I worked out that the example in my previous comment wasn't the best, as only lookarounds or backref validity checks can be used in the conditional expression.

I was wrong again :sweat_smile: but it's not clear to me how it should work with the IX if the condition consumes some characters (i.e. is a non-lookaround) and fails halfway through. Does the false branch execute from that position, or does it need to get reset to what it was before the condition was evaluated? I couldn't work it out from the Oniguruma test cases.

robinst: Hmm would this be enough for a test of that behavior?: `(?(ab)c|d)`. It matches only `d` in `ad`, not `ad`.

So that would suggest the second option, it is reset. That's also what I would have expected after reading the info page and before trying it out.

keith-hall: Thanks for checking. :) So we could treat it as an implicit lookahead then? This would match what RegexBuddy was implying. I'll try that when I get time.

keith-hall: Or we only need to reset when the false branch is hit? I guess I could also try Oniguruma locally with some additional test cases for exploration purposes. Especially as RegexBuddy doesn't seem to have an engine/mode specifically for Oniguruma.

keith-hall: Silly question, how are you executing arbitrary regex patterns against Oniguruma?

None of [the samples on their repo](https://github.com/kkos/oniguruma#sample-programs) seem to allow passing in a pattern and text as arguments on the command line, surely I'm missing something simple...

robinst: I normally use the rust onig crate's API. I think they also have examples that allow you to execute patterns: https://github.com/rust-onig/rust-onig

robinst: (BTW FYI I'm going to be on holidays for some time, so might be slow to respond :))

keith-hall: Thanks for the tip - that helped a lot :) I have pushed some more changes and test cases which helps align with Oniguruma further. Have a great holiday! I'm in no hurry for feedback, so please enjoy/relax and don't worry about this PR :)

src/parse.rs
```diff
@@ -679,6 +686,74 @@ impl<'a> Parser<'a> {
...
+        let mut if_false: Expr = Expr::Empty;
+        if let Expr::Alt(mut alternatives) = child {
+            // the truth branch will be the first alternative
+            if_true = alternatives.remove(0);
+            // if there is only one alternative left, take it out the Expr::Alt
+            if alternatives.len() == 1 {
+                if_false = alternatives.pop().expect("expected 2 alternatives");
+            } else {
+                // otherwise the remaining branches become the false branch
+                if_false = Expr::Alt(alternatives);
```
robinst, line 731: Nice! :)

src/compile.rs
```diff
@@ -210,6 +213,38 @@ impl Compiler {
         Ok(())
     }
 
+    fn compile_conditional<F>(&mut self, mut handle_child: F) -> Result<()>
+    where
+        F: FnMut(&mut Compiler, usize) -> Result<()>,
+    {
+        self.b.add(Insn::BeginAtomic);
```
robinst, line 220: It's kinda explained in the `EndAtomic` below, but it would be nice to have a comment here explaining why we need to use atomic here.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 with the exception that 0.x versions can break between minor versions.
 
 ## Unreleased
+### Added
+- Support for [conditionals](https://www.regular-expressions.info/conditional.html).
```
robinst, line 11: Can you add a small example (regex plus some text it matches against) here that gives an idea of what this is about?

robinst (approved):
Had a chance to look at this now, looks very good now, well done! You just need to resolve conflicts (sorry) and address the minor comments then we can merge.

CHANGELOG.md
keith-hall: good idea, done. I somehow missed that we had done that for previously added features, sorry :)

robinst:
Thanks, great work! Merged now.

robinst:
Released in version 0.11.0 now 🎉: https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md#0110---2023-01-12

PR #111, opened 2023-07-31
Add no-std support

Systemcluster:
[The `regex` crate added `no_std` support in a recent release](https://github.com/rust-lang/regex/blob/master/CHANGELOG.md#190-2023-07-05). This PR adds `no_std` support to `fancy_regex`, which mostly consists of just changing some imports from `std::` to their `core::` and `alloc::` counterparts.

~~The largest change is switching from the standard `HashMap` to the one provided by `hashbrown`.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "fancy-regex"
-version = "0.11.0" # remember to update html_root_url
+# remember to update html_root_url
+version = "0.11.0"
 authors = ["Raph Levien <raph@google.com>", "Robin Stocker <robin@nibor.org>"]
 edition = "2018"
 license = "MIT"
@@ -12,28 +13,34 @@ categories = ["text-processing"]
 exclude = ["/.github/*", "/Cargo.lock.msrv"]
 
 [features]
-default = ["unicode", "perf"]
+default = ["unicode", "perf", "std"]
 # Enable #[track_caller] in unit tests.
 track_caller = []
+
+std = ["regex/std"]
+
 perf = ["regex/perf"]
 perf-dfa = ["regex/perf-dfa"]
+perf-onepass = ["regex/perf-onepass"]
+perf-backtrack = ["regex/perf-backtrack"]
 perf-inline = ["regex/perf-inline"]
 perf-literal = ["regex/perf-literal"]
 perf-cache = ["regex/perf-cache"]
+
 unicode = ["regex/unicode"]
 
 [dependencies.regex]
-version = "1.3.8"
+version = "1.9.1"
 default-features = false
-features = ["std"]
 
-[dependencies]
-bit-set = "0.5"
+[dependencies.bit-set]
+version = "0.5"
+default-features = false
 
 [dev-dependencies]
-criterion = "0.3.0"
-matches = "0.1.8"
-quickcheck = "1.0.1"
+criterion = "0.5.1"
+matches = "0.1.10"
+quickcheck = "1.0.3"
 
 [[bench]]
 name = "bench"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -12,28 +13,37 @@ categories = ["text-processing"]
 exclude = ["/.github/*", "/Cargo.lock.msrv"]
 
 [features]
-default = ["unicode", "perf"]
+default = ["unicode", "perf", "std"]
```
BurntSushi, line 16: Do note that this is unfortunately a breaking change.

Cargo.toml
```diff
@@ -12,28 +13,37 @@ categories = ["text-processing"]
...
+default-features = false
+
+[dependencies.bit-set]
+version = "0.5"
 default-features = false
-features = ["std"]
 
 [dependencies]
-bit-set = "0.5"
+hashbrown = "0.14.0"
```
BurntSushi, line 41: Have you considered just using a `BTreeMap` when `HashMap` is unavailable instead of adding a new unconditional dependency? This is what the `regex` crate itself does.

Systemcluster: Seems `BTreeMap` is actually faster than `HashMap` by around 4% in almost all the included benchmarks on my machine (except for `run_backtrack_limit`, where it is 12% slower). Change made.

Cargo.toml
robinst: Thanks. I'm fine with this as we're still on 0.x.

src/expand.rs
```diff
@@ -128,45 +129,43 @@ impl Expander {
...
-        *dst = String::from_utf8(cursor.into_inner()).expect("expansion is UTF-8");
+        *dst = String::from_utf8(cursor).expect("expansion is UTF-8");
     }
 
     /// Writes the expansion produced by `expansion` to `dst`.  Potentially more efficient
     /// than calling `expansion` directly and writing the result.
     pub fn write_expansion(
         &self,
-        mut dst: impl io::Write,
+        dst: &mut Vec<u8>,
```
robinst, line 151: This is another breaking change. @johnw42 any opinion on this?

robinst (changes requested):
Thanks, changes look good.

Looks like we need to bump MSRV (minimum supported Rust version). Let's bump to 1.61.0 (the regex crate is on 1.60 but memchr requires 1.61)? Can you bump it in `.github/workflows/ci.yml`?

src/expand.rs
Systemcluster: When I implemented this I was sure this wasn't in the public API, but I must have overlooked something. I'll gladly change this. I'm seeing multiple options here:

- Add a new function `write_expansion_vec` and gate the current one behind `cfg(feature = "std")`
- Add another `write_expansion` and activate either based on the feature
- Write a trait and implement it for either `T: std::io::Write` or `Vec<u8>` based on the feature and similarly use a type alias for the returned `Result`

I added a commit with the first option. Let me know if I should change anything.

robinst (approved):
Thanks for your patience. Looks good now, going ahead and merging this.

Systemcluster:
Thanks for merging!

It looks like the MSRV check fail after the merge because `ansstyle`, a subdependency of `criterion`, raised the MSRV in a patch release between `1.0.2` and `1.0.3`. Relevant criterion issue here: https://github.com/bheisler/criterion.rs/pull/731
Is this just a matter of having to regenerate the `Cargo.lock.msrv`?

robinst:
Hey, thanks for checking. Yeah I had to update `Cargo.lock.msrv`, see:
* https://github.com/fancy-regex/fancy-regex/pull/118

robinst:
Released in version 0.12.0 🎉 : https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md

PR #116, opened 2023-10-25
Make Error and its components Clone

ch3pjw:
This brings it to parity with the underlying [regex](https://docs.rs/regex/latest/src/regex/error.rs.html#8-32) library


keith-hall (approved):

robinst:
Thank you! And thank you @keith-hall for reviewing :)

PR #117, opened 2023-11-04
Relative backref, assertions in character classes, anchored search, serializable ExprTree, etc

ruihe774:
Hi. I have been happy patching fancy-regex in recent days. Now I have a somewhat large patchset. I am wondering whether you are interested in it and whether it could be merged into upstream in batch. I can split my PR if you are not satisfied with a large PR.

I have made the following changes, some of which are semver breaking:

- I have added support for relative backref (`\g`). This allows you to specify absolute backrefs (e.g., `\g1`, `\g<name>`, same as `\k`) and relative backrefs (e.g., `\g<-1>`). Oniguruma also supports the `\g'name'` syntax, but I have only added support for `\g<name>`.

- `Regex` now caches a `Vec` for `Captures`. This is to avoid allocating a `Vec` for storing capture locations every time the user calls relevant APIs. I have used a spin lock to store the cache internally in `Regex`. Only `try_lock` is used, and a new `Vec` is allocated as a fallback. It has almost zero cost. Usually, the user only owns one `Captures`, so the cache can hit most of the time.
  This is not the only way to avoid allocation. The `regex` crate uses an explicit `CaptureLocations` and lets the user reuse it. This approach requires introducing new APIs.

- I have added support for assertions in character classes. Currently, we delegate character classes (e.g., `[a-z]`) to `regex`. Oniguruma supports assertions like `\A\z\b\<\>` in character classes, which `regex` does not support. I have added extra handling of such classes by wrapping the class in `Alt` or `Conditional`. However, nested classes are not supported.

- I now use `regex-automata` to perform anchored search. Previously, we used `inner` and `inner1` regexes in `Delegate` to perform non-lookleft and lookleft matches, respectively. This is because `^` only matches the start of text but not the start of the range. I have switched to using `regex-automata`, which can perform anchored search from the start of the range while having contexts out of range. Lookleft analysis is also dropped.

- I use `regex-syntax` to build the delegate regex. Previously, we rebuilt the regex string for the delegated subtree and let `regex` reparse it again. It was inefficient. I have implemented `Expr::to_ast()` and `Expr::to_hir()` to convert our `Expr` to `Ast` and further to `Hir` directly. `regex-automata` is then built from `Hir` directly during compilation.

- I have made `ExprTree` serializable/deserializable (when the "serde" feature is enabled). This can eliminate the parsing cost for prebuilt regexes. I have added a method `RegexBuilder::build_from_expr_tree` to build `Regex` directly from `ExprTree`, and a method `Regex::as_expr_tree` to export the `ExprTree`. Meanwhile, the `RegexImpl` is unified and removed, since the `Wrap` situation can be easily compiled to the whole delegate.

- I have added `Regex::captures_within_range`. This allows correct handling of word boundary matches at the end of the range.

- I have added optimization steps, including literal folding and flattening, in `Parser` to allow efficient serialization of `ExprTree` (otherwise char by char) and efficient compilation and `Insn` generation.

- I have implemented a case-insensitive `Lit` and a heuristic literal const size checking.

- I have implemented `Assertion` in `Expr` and `Insn` for assertions like `StartLine` and `WordBoundary`. It is implemented by the VM now. (Previously we just delegated it.)

This PR is marked as WIP because I haven't written and updated tests and docs. Some tests are temporarily disabled due to changes in internal representations. If you are interested in this PR, I would like to add tests and docs in the near future.


ruihe774:
BTW, I have also implemented a prefilter (the optimization you've mentioned in PERFORMANCE.md) in [prefilter.rs](https://github.com/ruihe774/fancy-regex/blob/msk/src/prefilter.rs). In my end-to-end benchmarking (syntect source code highlighting using fancy-regex), the performance improves significantly (~20%-60%). I'm wondering if you are interested in it.

keith-hall:
Wow, nice, lots of useful changes here, thanks for submitting the PR.

I'm no Rust expert but I'll try to review what I can when I get some free time, maybe I'll have some questions which could save time for other reviewers.

> * I have added support for assertions in character classes. Currently, we delegate character classes (e.g., `[a-z]`) to `regex`. Oniguruma supports assertions like `\A\z\b\<\>` in character classes

Interesting, I don't see that mentioned in the Oniguruma docs, nor have I ever seen it in the wild.
https://github.com/kkos/oniguruma/blob/master/doc/RE#L205-L235

> BTW, I have also implemented a prefilter (the optimization you've mentioned in PERFORMANCE.md) in [prefilter.rs](https://github.com/ruihe774/fancy-regex/blob/msk/src/prefilter.rs). In my end-to-end benchmarking (syntect source code highlighting using fancy-regex), the performance improves significantly (~20%-60%). I'm wondering if you are interested in it.

I personally think we'd be crazy to not be interested, especially with such significant speedups, but that's just my vote - I can't wait to see the other maintainers reactions (and hope I haven't just called them crazy lol 😅)

ruihe774:
The CI failed. Maybe it is because the MSRV.

>> I have added support for assertions in character classes. Currently, we delegate character classes (e.g., [a-z]) to regex. Oniguruma supports assertions like \A\z\b\<\> in character classes
>
> Interesting, I don't see that mentioned in the Oniguruma docs, nor have I ever seen it in the wild.
https://github.com/kkos/oniguruma/blob/master/doc/RE#L205-L235

Sorry, I misunderstood that. I've checked with Oniguruma again and it turns out that it just ignores such escapes. So `[\<\>]` will just literally match `<` and `>`. `[\<\>]` will make `regex` error, so what we need to do is just simply workaround it in this case. 😂

> Regex now caches a Vec for Captures. This is to avoid allocating a Vec for storing capture locations every time the user calls relevant APIs. I have used a spin lock to store the cache internally in Regex. Only try_lock is used, and a new Vec is allocated as a fallback. It has almost zero cost. Usually, the user only owns one Captures, so the cache can hit most of the time.

I think I need to discuss this change more. I now find that majority of allocations, instead of captures, is actually the `State` of VM. The `Vec` for stack and saves will allocate and deallocate every time. I think we need to avoid that.

There are three approaches:
- Use a `SpinMutex` to cache `State`. Allocating it only if multiple threads are running the VM.
- Use a `RefCell` to store `State` and make `Regex` single-threaded (i.e. lose `Sync` trait). This is what I have done on [my branch](http://github.com/ruihe774/fancy-regex/tree/msk) because I know my application is single-threaded.
- Wrap the `State` and let the user to cache and reuse it. We can provide a set of advanced APIs to let user pass their cached `State` to perform search.

The reason why I want to eliminate allocations in matching is that:
- Heap allocation can be slow. For example, the system allocator of newer macOS, which Rust uses, will zero the memory on deallocation.
- Matching without allocation can be made able to run with `core` and no `std`. The user can skip the parsing and compiling, and store `Prog` to use regex. However, providing such a set of no-`std` APIs is out of current topic.

As an example, `regex` crate and underlying `regex-automata` crate do no allocation in matching and provide a no-`std` and no-`alloc` feature.

> I personally think we'd be crazy to not be interested, especially with such significant speedups, but that's just my vote - I can't wait to see the other maintainers reactions (and hope I haven't just called them crazy lol 😅)

You can have a look at my [recent commits](https://github.com/ruihe774/fancy-regex/commits/msk) (c24bbe6 and later) to see what I have done. I don't think it's reasonable to present it in this PR, though, as that will make the size of this PR grow crazy. We can do things progressively 😸

ruihe774:
The new MSRV is 1.66, because we need the feature "mixed_integer_ops"

keith-hall:
I have approved the updated workflow 👍

> I think I need to discuss this change more.

Unfortunately I don't feel like I'm the best person to discuss this with, I'm rather hoping @robinst will be able to find some time to join in this discussion soon 🙂
That said, I certainly agree with the idea to remove allocations from matching where possible, that reasoning is sound

> I don't think it's reasonable to present it in this PR, though, as that will make the size of this PR grow crazy. We can do things progressively 😸

Totally no objections from me, that seems sensible!

ruihe774:
> There are three approaches: ...

Ok, I found and switched to use [`regex_automata::util::pool::Pool`](https://docs.rs/regex-automata/latest/regex_automata/util/pool/struct.Pool.html). Great!

robinst:
> Unfortunately I don't feel like I'm the best person to discuss this with, I'm rather hoping @robinst will be able to find some time to join in this discussion soon 🙂

Hi :). Thanks for your patience. What an epic PR!

I've just merged the other open PRs and released a new version. Unfortunately it means this PR now has conflicts. Can you resolve them by rebasing?

I think all the changes sound great, but I haven't had a chance to look through all the code yet. It would be a big help if you could split out some of the changes into separate PRs, especially the new APIs and optimizations. I hope it's not too much effort to do that?

> Ok, I found and switched to use [`regex_automata::util::pool::Pool`](https://docs.rs/regex-automata/latest/regex_automata/util/pool/struct.Pool.html). Great!

Great! The more we can reuse from `regex` the better :). When I read https://blog.burntsushi.net/regex-internals/ I was thinking it would be cool if we could use some of it for fancy-regex, and it looks like that's exactly what you've done, so good!

ruihe774:
> I've just merged the other open PRs and released a new version. Unfortunately it means this PR now has conflicts. Can you resolve them by rebasing?

Yes. I'm working on this.

> I think all the changes sound great, but I haven't had a chance to look through all the code yet. It would be a big help if you could split out some of the changes into separate PRs, especially the new APIs and optimizations. I hope it's not too much effort to do that?

I've squashed commits of each change into a single commit. I think I can update the CHANGLOG.md and specify all my changes and the corresponding commit for each of them, so that you can review them one commit by commit. Does this approach sound reasonable for you?

robinst:
> I've squashed commits of each change into a single commit. I think I can update the CHANGLOG.md and specify all my changes and the corresponding commit for each of them, so that you can review them one commit by commit. Does this approach sound reasonable for you?

You mean the current 23 commits? It's a lot of commits, and I see there's some things that later commits change again. That seems very hard to review commit by commit. I'd much prefer 4-5 self-contained pull requests. (I'm also not sure how GitHub's support for leaving comments on individual commits is, it seems like it could get pretty messy. But that's a separate issue.)

ruihe774:
> and I see there's some things that later commits change again. That seems very hard to review commit by commit.

Yep, that's the issue. Some later commits that modify or fix previous changes may depend on commits for other changes, and cherry picking them back may be difficult. Anyway, I will try my best to separate them. For changes that have dependencies, I will open PRs for them after their precedent PRs get merged.

PR #121, opened 2023-11-11
Part 1: regex-automata, relative backref, escaping inhibition in classes, Expr::Assertion, Lit { casei }

ruihe774:
This PR contains the first part of #117:

- Switch to use `regex-syntax` and `regex-automata` for anchored delegated search.
- Support `\g` and relative backrefs.
- Inhibit escaping in character classes.
- Implement assertions in VM.
- ~Implement case-insensitive Lit in VM.~

Sorry but I cannot break down it any further. Conflicts would eat me🤣. And a huge number of diff line count is caused by the moving of `Expr` into `parse.rs`, so this is actually a PR with less than 500 lines added. 

Some tests are disabled temporarily, which will be updated and re-enabled in later PR.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -590,7 +590,6 @@ mod tests {
     use crate::vm::Insn::*;
     use alloc::vec;
     use bit_set::BitSet;
-    use matches::assert_matches;
 
     #[test]
     fn jumps_for_alternation() {
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -578,6 +578,7 @@ impl DelegateBuilder {
     }
 }
 
+#[cfg(not(test))]
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -569,16 +569,12 @@ impl DelegateBuilder {
         let end_group = self.end_group;
 
         let compiled = compile_inner(&self.re, options)?;
-        if self.const_size && start_group == end_group {
-            let size = self.min_size;
-            Ok(Insn::DelegateSized(compiled, size))
-        } else {
-            Ok(Insn::Delegate {
-                inner: compiled,
-                start_group,
-                end_group,
-            })
-        }
+
+        Ok(Insn::Delegate {
+            inner: compiled,
+            start_group,
+            end_group,
+        })
     }
 }
 
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -729,5 +729,4 @@ mod tests {
 
     #[cfg(not(feature = "std"))]
     fn assert_delegate(_: &Insn, _: &str) {}
-
 }
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -578,7 +578,6 @@ impl DelegateBuilder {
     }
 }
 
-#[cfg(not(test))]
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -587,6 +586,7 @@ mod tests {
     use crate::vm::Insn::*;
     use alloc::vec;
     use bit_set::BitSet;
+    use matches::assert_matches;
 
     #[test]
     fn jumps_for_alternation() {
@@ -663,7 +663,7 @@ mod tests {
         assert_matches!(prog[0], Split(1, 3));
         assert_matches!(prog[1], Lit(ref l) if l == "x");
         assert_matches!(prog[2], FailNegativeLookAround);
-        assert_delegate_sized(&prog[3], "(?:a|b)c");
+        assert_delegate(&prog[3], "(?:a|b)c");
         assert_delegate(&prog[4], "x*");
         assert_matches!(prog[5], End);
     }
@@ -730,25 +730,4 @@ mod tests {
     #[cfg(not(feature = "std"))]
     fn assert_delegate(_: &Insn, _: &str) {}
 
-    #[cfg(feature = "std")]
-    fn assert_delegate_sized(insn: &Insn, re: &str) {
-        match insn {
-            Insn::DelegateSized(inner, ..) => {
-                assert_eq!(
-                    PATTERN_MAPPING
-                        .read()
-                        .unwrap()
-                        .get(&alloc::format!("{:?}", inner))
-                        .unwrap(),
-                    re
-                );
-            }
-            _ => {
-                panic!("Expected Insn::DelegateSized but was {:#?}", insn);
-            }
-        }
-    }
-
-    #[cfg(not(feature = "std"))]
-    fn assert_delegate_sized(_: &Insn, _: &str) {}
 }
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -20,9 +20,13 @@
 
 //! Compilation of regexes to VM.
 
-use alloc::boxed::Box;
-use alloc::string::{String, ToString};
+use alloc::string::String;
 use alloc::vec::Vec;
+use core::usize;
+use regex_automata::meta::Regex as RaRegex;
+use regex_automata::meta::{Builder as RaBuilder, Config as RaConfig};
+#[cfg(all(test, feature = "std"))]
+use std::{collections::BTreeMap, sync::RwLock};
 
 use crate::analyze::Info;
 use crate::vm::{Insn, Prog};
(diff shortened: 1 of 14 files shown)


src/compile.rs
```diff
@@ -444,14 +445,22 @@ impl Compiler {
...
-        // TODO: might want to do something similar for case insensitive literals
-        // (have is_literal return an additional bool for casei)
-        if infos.iter().all(|e| e.is_literal()) {
-            let mut val = String::new();
+
+        if infos
+            .iter()
+            .all(|info| matches!(info.expr, Expr::Literal { .. }))
+        {
+            // Parser::optimize() will fold literals, so infos.len() is usually 1.
```
ruihe774, line 453: Well, that will reside in another PR...

src/parse.rs
```diff
@@ -38,16 +44,281 @@ const FLAG_SWAP_GREED: u32 = 1 << 3;
 const FLAG_IGNORE_SPACE: u32 = 1 << 4;
 const FLAG_UNICODE: u32 = 1 << 5;
 
-#[cfg(not(feature = "std"))]
-pub(crate) type NamedGroups = alloc::collections::BTreeMap<String, usize>;
-#[cfg(feature = "std")]
-pub(crate) type NamedGroups = std::collections::HashMap<String, usize>;
+/// Regular expression AST. This is public for now but may change.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
ruihe774, line 49: This is actually for serde support (which will be in another PR). Just ignore it and it will not cause any issues for now.

src/parse.rs
```diff
@@ -38,16 +44,281 @@ const FLAG_SWAP_GREED: u32 = 1 << 3;
...
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Assertion {
+    /// Start of input text
+    StartText,
+    /// End of input text
+    EndText,
+    /// Start of a line
+    StartLine {
+        /// CRLF mode
+        crlf: bool,
```
ruihe774, line 176: Not implemented in parser for now.

Cargo.lock.direct-minimal-versions
ruihe774, line 1: This is generated using [direct-minimal-versions](https://doc.rust-lang.org/cargo/reference/unstable.html#direct-minimal-versions).

ruihe774, line 1: This is generated using [direct-minimal-versions](https://doc.rust-lang.org/cargo/reference/unstable.html#direct-minimal-versions).

Cargo.lock.minimal-versions
ruihe774, line 1: This is generated by [minimal-versions](https://doc.rust-lang.org/cargo/reference/unstable.html#minimal-versions).

src/vm.rs
```diff
@@ -484,12 +487,62 @@ pub(crate) fn run(
...
+                        }
+                    } else {
+                        if !eq_test_exact_size(sb.iter().copied(), val.bytes(), |a, b| {
+                            // Do we need a == b shortcut? Is to_ascii_lowercase() fast?
+                            a == b || a.to_ascii_lowercase() == b.to_ascii_lowercase()
+                        }) {
+                            if !s.get(ix..end).map_or(false, |s| {
+                                eq_test(
+                                    s.chars().flat_map(char::to_lowercase),
+                                    val.chars().flat_map(char::to_lowercase),
```
robinst, line 537: Was this copied from somewhere? Case-insensitive matching is a big can of worms. I'm not sure we should try to do it in fancy-regex. I'd prefer that we keep delegating these to regex.

tests/oniguruma.rs
```diff
@@ -220,11 +213,7 @@ fn oniguruma() {
             assert!(result.is_some(),
                     "Expected ignored test to fail, but it succeeded. Remove it from the ignore file: {}", &test.source);
             let failure = result.unwrap();
-            // rustc before version 1.53 unnecessarily escapes slashes in debug format strings
-            // meaning when run on our MSRV, we get different failure text than on the latest stable Rust
-            // so here we try removing the backslash when comparing. After our MSRV is bumped up to
-            // 1.53 or higher, we can remove this part again.
```
robinst, line 226: Nice, thanks.

robinst (changes requested):
Did a first pass, added some comments.

src/parse.rs
```diff
@@ -38,16 +44,279 @@ const FLAG_SWAP_GREED: u32 = 1 << 3;
 const FLAG_IGNORE_SPACE: u32 = 1 << 4;
 const FLAG_UNICODE: u32 = 1 << 5;
 
-#[cfg(not(feature = "std"))]
-pub(crate) type NamedGroups = alloc::collections::BTreeMap<String, usize>;
-#[cfg(feature = "std")]
-pub(crate) type NamedGroups = std::collections::HashMap<String, usize>;
```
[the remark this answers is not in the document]
robinst: Not sure why you marked this as resolved, still looks like the same change?

src/parse.rs
```diff
@@ -839,7 +1134,7 @@ pub(crate) fn parse_id<'a>(s: &'a str, open: &'_ str, close: &'_ str) -> Option<
 }
 
 fn is_id_char(c: char) -> bool {
-    c.is_alphanumeric() || c == '_'
+    c.is_alphanumeric() || c == '_' || c == '-'
```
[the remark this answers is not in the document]
robinst: Can you provide an explanation?

src/vm.rs
robinst: I'd prefer not to include this in this PR. It would need a lot of tests and as I said before, I'd rather just delegate this to regex for correctness. If it's in a separate PR we can look at benchmarks too.

robinst (changes requested):
Thanks, much better to review! Added some comment to unresolved parts. Once those are addressed we can hopefully merge this and get to the next one :).

src/parse.rs
```diff
@@ -839,7 +867,7 @@ pub(crate) fn parse_id<'a>(s: &'a str, open: &'_ str, close: &'_ str) -> Option<
...
```
robinst, line 870: Note that both Ruby and Python don't allow `-` in group names:

```
SyntaxError ((irb):14: invalid group name <foo-bar>>: /(?<foo-bar>\w+)%\k<foo-bar>/)
```

```
re.error: bad character in group name 'foo-bar' at position 4
```

But I now see that the reason you're doing this change is so that you can parse `\k<-1>`. The code needs to be changed around a bit. Can you add tests for these:

* `\k<foo-bar>` -> invalid group name
* `\k<-1>` -> works as a relative backref
* `\k<+1>` -> works as a relative backref (different from `\k<1>`)

To support that, I'd probably just check the first character within `parse_named_backref`. If it's a `-` or `+`, parse the rest as a number. Otherwise, parse the rest with `c.is_alphanumeric() || c == '_'`.

src/vm.rs
```diff
@@ -484,12 +487,62 @@ pub(crate) fn run(
...
+                }
+                Insn::Lit { ref val, casei } => {
+                    let end = ix.saturating_add(val.len()).min(s.len());
+                    let sb = &s.as_bytes()[ix..end];
+                    if !casei {
+                        if sb != val.as_bytes() {
+                            break 'fail;
+                        }
+                    } else {
+                        if !eq_test_exact_size(sb.iter().copied(), val.bytes(), |a, b| {
```
robinst, line 530: You removed the other part of this code (in `visit`) but please also remove this part for now.

src/parse.rs
```diff
@@ -38,10 +41,7 @@ const FLAG_SWAP_GREED: u32 = 1 << 3;
 const FLAG_IGNORE_SPACE: u32 = 1 << 4;
 const FLAG_UNICODE: u32 = 1 << 5;
 
-#[cfg(not(feature = "std"))]
-pub(crate) type NamedGroups = alloc::collections::BTreeMap<String, usize>;
-#[cfg(feature = "std")]
-pub(crate) type NamedGroups = std::collections::HashMap<String, usize>;
+pub(crate) type NamedGroups = BTreeMap<String, usize>;
```
robinst, line 44: Please undo this change, as far as I can tell it doesn't have anything to do with the PR.

robinst (changes requested):

robinst (approved):
@ruihe774 I've pushed a few commits to address some things and fix the build, please have a look. If that all looks good, we can merge :)

ruihe774:
LGTM👍

robinst:
Merged! Thanks, looking forward to the next one. Try to keep it smaller than this please :).

robinst:
@ruihe774 Hey, I was just writing the release notes to release this but I noticed that `\g` is not just another syntax for backrefs, it's subexp calls (see https://www.regular-expressions.info/subroutine.html and https://github.com/kkos/oniguruma/blob/master/doc/RE#L460).

I'll revert the `g` part of the change in a new PR (but keeping the relative change, i.e. `\k<-1>`).

ruihe774:
> @ruihe774 Hey, I was just writing the release notes to release this but I noticed that `\g` is not just another syntax for backrefs, it's subexp calls (see https://www.regular-expressions.info/subroutine.html and https://github.com/kkos/oniguruma/blob/master/doc/RE#L460).
> 
> I'll revert the `g` part of the change in a new PR (but keeping the relative change, i.e. `\k<-1>`).

Interesting. So we need a separated implementation for `\g`. I'd like to look into this latter, maybe.

PR #128, opened 2023-12-19
Undo parsing \g as backref

robinst:
See:
* https://github.com/fancy-regex/fancy-regex/pull/121#issuecomment-1857255661


robinst:
CC @ruihe774

PR #131, opened 2024-01-03
Introduce `R` opt-in flag for the CRLF mode

mmizutani:
## Description

This PR introduces the inline regex flag `R` (`(?R:)`) for switching on and off the CRLF mode.

## Changes

- Extended the parser to support a new opt-in regex flag `R` for switching the CRLF mode. 
- Extended the backtracking VM engine to support the CRLF mode.

## Background

The CRLF mode has been available in the Rust `regex` crate via an opt-in flag since [v1.9.0](https://github.com/rust-lang/regex/blob/master/CHANGELOG.md#190-2023-07-05) and in the corresponding versions of the `regex-automata` crate, 

> A new inline flag, `R`, which enables CRLF mode. This makes `.` match any Unicode scalar value except for `\r` and `\n`, and also makes `(?m:^)` and `(?m:$)` match after and before both `\r` and `\n`, respectively, but never between a `\r` and `\n`.

but it is still not supported in the `fancy-regex` crate:

> regex must be valid: ParseError(2, UnknownFlag("(?R"))

The CRLF mode was introduced to the Rust `regex` crate to make the treatment of a carriage return `r` by any character or byte (`.`) and line anchors (`^`, `$`) more aligned with that of a line feed `\n` in a backward-compatible way.

Before the introduction of the CRLF mode or with the CRLF mode turned off, which is by default in the `regex` crate, a carriage return `r` is not fully handled as a new line character unlike `\n` and is treated more like an ordinary, non- new line character:

- When the multiline mode is on and the state (dot new line) mode is off (`(?m-s)`), the any syntax (`.`) matches ~~`\n`~~`\r` but not ~~`\r`~~`\n` in the Rust `regex` crate engine as if `. ＝ [^\n]` not `. ＝ [^\r\n]`. This behavior is [in line with default behavior of other major regex engines](https://github.com/BurntSushi/rebar/blob/11a62af1c9175654aa8bdf4135fc2912bc36c3ff/benchmarks/definitions/test/dot.toml#L133-L196), but some use cases might prefer `\r` being treated as a new line character just like `\n`.
- Line anchors (`^` and `$` in the multiline mode (`(?m:)`) could find a match at a boundary between `\r` and `\n` of `\r\n` in the rust `regex` crate engine although it is not an end of a line in a strict sense. In Oniguruma and some other regex engines, `^` and `$` do not match at the middle of `\r\n`, or how to treat `\r`, `\n`, and other less common new line Unicode characters is configurable as in PCRE2.

Being able to selectively enable the CRLF mode in the `fancy-regex` would be useful in some special cases, e.g., when porting regex strings written for the PCRE2 engine compiled with the `PCRE2_NEWLINE_ANYCRLF` option.

## Note

~~The CRLF mode is a bit tricky to understand since the `R` inline flag indirectly changes matching behaviors by altering the effects of two other inline flags: multiline (`m`) and state (dot new line) (`s`).
In a nutshell,~~

As BurntSushi has pointed out in the comment below, we can think about CRLF mode as changing a single thing: the line terminator goes from `\n` to `\r\n|\r|\n`. Everything else follows from that, including the behavior of the `s` and `m` flags.

Specifically, from the perspective of the developers of `fancy-regex`, this implies the following branching of behaviors of `^`, `$` and `.` based on the presence or absence of the `R` flag:

| Flags | `^` and `$` behave as | Effects on matching  |
| --- | --- | --- |
| `(?)`, `(?R)` | `Assertion::StartText`, <br>`Assertion::EndText` | `^` and `$` match the start and end of the whole input text. That is, `R` flag makes no difference. |
| `(?m)` | `Assertion::StartLine { crlf: false }`,  <br>`Assertion::EndLine { crlf: false }` | `^` and `$` match starts and end of lines. They could match at between `\r` and `\n` of `\r\n`. |
| `(?mR)` | `Assertion::StartLine { crlf: true }`, <br>`Assertion::EndLine { crlf: true }` | `^` and `$` match start and end of lines. They never match at between `\r` and `\n` of `\r\n`. `\r\n` as a whole is treated like a newline. |

| Flags |  `.` behaves as | Effects on matching |
| --- | --- | --- |
| `(?s)`, `(?sR)` | `Insn::Any` | `.` matches any character including `\r` and `\n`. That is, `R` flag makes no difference. |
| `(?)` | `Insn::AnyExceptLF` | `.` matches any character including `\r` but excluding `\n`.  |
| `(?R)` | `Insn::AnyExceptCRLF` | `.` matches any character excluding `\r` and `\n`.  |

## Related discussion and commits

- $ doesn't match CRLF · Issue #244 · rust-lang/regex: https://github.com/rust-lang/regex/issues/244#issuecomment-1451258260
- syntax: add support for CRLF-aware line anchors · rust-lang/regex@b6085a9: https://github.com/rust-lang/regex/commit/b6085a96c246c61b8ad41ec5e2bfb95b52618943

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1363,12 +1363,16 @@ pub enum Assertion {
     EndText,
     /// Start of a line
     StartLine {
-        /// CRLF mode
+        /// CRLF mode.
+        /// If true, this assertion matches at the starting position of the input text, or at the position immediately
+        /// following either a `\r` or `\n` character, but never after a `\r` when a `\n` follows.
         crlf: bool,
     },
     /// End of a line
     EndLine {
-        /// CRLF mode
+        /// CRLF mode.
+        /// If true, this assertion matches at the starting position of the input text, or at the position immediately
+        /// following either a `\r` or `\n` character, but never after a `\r` when a `\n` follows.
         crlf: bool,
     },
     /// Left word boundary
@@ -1626,6 +1630,28 @@ mod tests {
     }
 
     #[test]
+    fn to_str_assertion() {
+        assert_eq!(to_str(Expr::Assertion(crate::Assertion::StartText)), "^");
+        assert_eq!(to_str(Expr::Assertion(crate::Assertion::EndText)), "$");
+        assert_eq!(
+            to_str(Expr::Assertion(crate::Assertion::StartLine { crlf: false })),
+            "(?m:^)"
+        );
+        assert_eq!(
+            to_str(Expr::Assertion(crate::Assertion::EndLine { crlf: false })),
+            "(?m:$)"
+        );
+        assert_eq!(
+            to_str(Expr::Assertion(crate::Assertion::StartLine { crlf: true })),
+            "(?Rm:^)"
+        );
+        assert_eq!(
+            to_str(Expr::Assertion(crate::Assertion::EndLine { crlf: true })),
+            "(?Rm:$)"
+        );
+    }
+
+    #[test]
     fn as_str_debug() {
         let s = r"(a+)b\1";
         let regex = Regex::new(s).unwrap();
@@ -1642,9 +1668,16 @@ mod tests {
 
     #[test]
     fn from_str() {
-        let s = r"(a+)b\1";
-        let regex = s.parse::<Regex>().unwrap();
-        assert_eq!(regex.as_str(), s);
+        for &s in &[
+            r"(a+)b\1",
+            r"(?m:^)",
+            r"(?m:$)",
+            r"(?Rm:^)",
+            r"(?Rm:$)",
+        ] {
+            let regex = s.parse::<Regex>().unwrap();
+            assert_eq!(regex.as_str(), s);
+        }
     }
 
     #[test]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -733,7 +733,10 @@ impl Regex {
     }
 
     fn new_options(options: RegexOptions) -> Result<Regex> {
-        let raw_tree = Expr::parse_tree_with_flags(&options.pattern, options.compute_flags())?;
+        let mut tree = Expr::parse_tree_with_flags(&options.pattern, options.compute_flags())?;
+
+        // try to optimize the expression tree
+        let requires_capture_group_fixup = optimize(&mut tree);
 
         // wrapper to search for re at arbitrary start position,
         // and to capture the match bounds
@@ -748,13 +751,11 @@ impl Regex {
                     hi: usize::MAX,
                     greedy: false,
                 },
-                Expr::Group(Box::new(raw_tree.expr)),
+                Expr::Group(Box::new(tree.expr)),
             ]),
-            ..raw_tree
+            ..tree
         };
 
-        // try to optimize the expression tree
-        let requires_capture_group_fixup = optimize(&mut tree);
         let info = analyze(&tree, if requires_capture_group_fixup { 0 } else { 1 })?;
 
         if !info.hard {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -504,7 +504,10 @@ impl Regex {
         let tree = ExprTree {
             expr: Expr::Concat(vec![
                 Expr::Repeat {
-                    child: Box::new(Expr::Any { newline: true, crlf: true }),
+                    child: Box::new(Expr::Any {
+                        newline: true,
+                        crlf: true,
+                    }),
                     lo: 0,
                     hi: usize::MAX,
                     greedy: false,
@@ -1413,10 +1416,22 @@ impl Expr {
     pub fn to_str(&self, buf: &mut String, precedence: u8) {
         match *self {
             Expr::Empty => (),
-            Expr::Any { newline: true, crlf: true } => buf.push_str("(?sR:.)"),
-            Expr::Any { newline: true, crlf: false } => buf.push_str("(?s:.)"),
-            Expr::Any { newline: false, crlf: true } => buf.push_str("(?R:.)"),
-            Expr::Any { newline: false, crlf: false } => buf.push_str("."),
+            Expr::Any {
+                newline: true,
+                crlf: true,
+            } => buf.push_str("(?sR:.)"),
+            Expr::Any {
+                newline: true,
+                crlf: false,
+            } => buf.push_str("(?s:.)"),
+            Expr::Any {
+                newline: false,
+                crlf: true,
+            } => buf.push_str("(?R:.)"),
+            Expr::Any {
+                newline: false,
+                crlf: false,
+            } => buf.push_str("."),
             Expr::Literal { ref val, casei } => {
                 if casei {
                     buf.push_str("(?i:");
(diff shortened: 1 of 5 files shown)


BurntSushi:
> When the multiline mode is on and the state (dot new line) mode is off (`(?m-s:)`), the any syntax (`.`) matches `\n` but not `\r` in the Rust `regex` crate engine as if `. ＝ [^\n]` not `. ＝ [^\r\n]`. 

To clarify, the _only_ way for `.` to match `\n` is when the `s` flag is enabled. The `m` flag has zero impact on the match semantics of `.`. I think your _wording_ is wrong, but your `. = [^\n]` is correct.

> In the roughly equivalent mode of Oniguruma (`(?-m:)`) and some other regex engines, `.` matches both `\n` and `\r`.

This doesn't seem right either:

```rust
fn main() {
    let regex = onig::Regex::new("(?-m:.)").unwrap();
    println!("{:?}", regex.find("\n"));
}
```

Has this output:

```
$ cargo run -q
None
```

But this:

```rust
fn main() {
    let regex = onig::Regex::new("(?m:.)").unwrap();
    println!("{:?}", regex.find("\n"));
}
```

Has this output:

```
$ cargo run -q
Some((0, 1))
```

Oniguruma is actually the odd regex engine here. [The vast majority of regex engines treat the `s` and `m` flags in the same way as the `regex` crate.](https://github.com/BurntSushi/rebar/blob/11a62af1c9175654aa8bdf4135fc2912bc36c3ff/benchmarks/definitions/test/dot.toml) I would not recommend copying Oniguruma on this point. I think Oniguruma might have gotten this behavior from Spencer's regex library, but I can't find a source for that.

> The CRLF mode is a bit tricky to understand since the R inline flag indirectly changes matching behaviors by altering the effects of two other inline flags: multiline (m) and state (dot new line) (s).

Think about CRLF mode as changing a single thing: the line terminator goes from `\n` to `\r\n|\r|\n`. Everything else follows from that, including the behavior of the `s` and `m` flags.

mmizutani:
@BurntSushi 
Thank you for your points as well as the comprehensive script for checking the behavior of various regex engines. I have corrected my PR description, especially,

> When the multiline mode is on and the state (dot new line) mode is off (`(?m-s)`), the any syntax (`.`) matches ~~`\n`~~`\r` but not ~~`\r`~~`\n` in the Rust `regex` crate engine as if `. ＝ [^\n]` not `. ＝ [^\r\n]`. This behavior is [in line with default behavior of other major regex engines](https://github.com/BurntSushi/rebar/blob/11a62af1c9175654aa8bdf4135fc2912bc36c3ff/benchmarks/definitions/test/dot.toml#L133-L196), but some use cases might prefer `\r` being treated as a new line character just like `\n` in a consistent way.

I have replaced Oniguruma with PCRE2 to reduce noise in explaining when the CRLF mode might be useful in `fancy-regex`. The naming of regex flags in Oniguruma is indeed rather unique and confusing for users of other regex engines.

src/lib.rs
```diff
@@ -1363,12 +1368,16 @@ pub enum Assertion {
...
+        /// CRLF mode.
+        /// If true, this assertion matches at the starting position of the input text, or at the position immediately
+        /// following either a `\r` or `\n` character, but never after a `\r` when a `\n` follows.
         crlf: bool,
     },
     /// End of a line
     EndLine {
-        /// CRLF mode
+        /// CRLF mode.
+        /// If true, this assertion matches at the starting position of the input text, or at the position immediately
```
keith-hall:
```suggestion
        /// If true, this assertion matches at the ending position of the input text, or at the position immediately
```

src/vm.rs
```diff
@@ -107,8 +107,10 @@ pub enum Insn {
     End,
     /// Match any character (including newline)
     Any,
-    /// Match any character (not including newline)
-    AnyNoNL,
+    /// Match any character (including a Line Feed (`\n`) but not including other newline characters)
```
keith-hall:
```suggestion
    /// Match any character excluding a Line Feed (`\n`) but not other newline characters
```

src/vm.rs
```diff
@@ -107,8 +107,10 @@ pub enum Insn {
...
+    AnyExceptLF,
+    /// Match any character (not including a Line Feed (`\n`) or a Carriage Return (`\r`)
```
keith-hall:
```suggestion
    /// Match any character excluding a Line Feed (`\n`) or a Carriage Return (`\r`)
```

src/parse.rs
```diff
@@ -39,6 +39,7 @@ const FLAG_DOTNL: u32 = 1 << 2;
 const FLAG_SWAP_GREED: u32 = 1 << 3;
 const FLAG_IGNORE_SPACE: u32 = 1 << 4;
 const FLAG_UNICODE: u32 = 1 << 5;
+const FLAG_CRLF: u32 = 1 << 6;
```
keith-hall: flags have been moved to a new file, `flags.rs` since this PR was made, please can you rebase/merge latest main branch? :)

keith-hall (commented):
Thanks for the PR and sorry you had to wait so long for a review! I admit I was a little hesitant about this one initially because AFAIK Oniguruma - which fancy-regex strives to be compatible with - doesn't offer such a flag. But I see that it [does have an escape char which serves a similar purpose](https://github.com/kkos/oniguruma/blob/master/doc/RE#L83-L87), plus it makes sense to expose the power of the regex crate where we can, so I think it should be fine to get this flag supported by fancy-regex.

src/lib.rs
```diff
@@ -1376,7 +1376,7 @@ pub enum Assertion {
     /// End of a line
     EndLine {
         /// CRLF mode.
-        /// If true, this assertion matches at the starting position of the input text, or at the position immediately
+        /// If true, this assertion matches at the ending position of the input text, or at the position immediately
         /// following either a `\r` or `\n` character, but never after a `\r` when a `\n` follows.
```
keith-hall: sorry, just noticed this one too:
```suggestion
        /// preceding either a `\r` or `\n` character, but never after a `\r` when a `\n` follows.
```

src/lib.rs
```diff
@@ -737,6 +738,25 @@ impl Regex {
 
         // try to optimize the expression tree
         let requires_capture_group_fixup = optimize(&mut tree);
+
+        // wrapper to search for re at arbitrary start position,
```
keith-hall, line 742: This wrapping is no longer needed btw, it was removed recently 🙂

src/parse.rs
```diff
@@ -579,7 +582,13 @@ impl<'a> Parser<'a> {
         } else if b == b'G' && !in_class {
             (end, Expr::ContinueFromPreviousMatchEnd)
         } else if b == b'O' && !in_class {
-            (end, Expr::Any { newline: true })
+            (
+                end,
+                Expr::Any {
+                    newline: self.flag(FLAG_DOTNL),
```
keith-hall, line 588: [According to Oniguruma docs, `\O` should always match newlines and carriage returns](https://github.com/kkos/oniguruma/blob/master/doc/RE#L91) 🙂

keith-hall:
superseded by https://github.com/fancy-regex/fancy-regex/pull/238

Viorel:
It is probably a good idea to have such a flag in **regex**, **regex-lite** and **fancy-regex**, but it should be mentioned that `(?R)` already exists in two major engines — Perl and PCRE2 — where `(?R)` has a different meaning. Therefore, expressions will be incompatible.

PR #132, opened 2024-01-24
RegexBuilder Options Port: Enable Overrides for RaBuilder

jonperry-dev:
# Case Insensitive Builder Feature Port

This is a small add-in to allow for options to be passed in as boolean values and mutate the `Config` through the `RegexBuilder`. This is primarily to align with the standard Regex crate for their own [RegexBuilder](https://docs.rs/regex/latest/src/regex/builders.rs.html#322-325).

# Changes

- Added a field to the `RegexOptions` to keep a state of `regex_automata::utils::syntax::Config` internally to disseminate the proper flag overrides
- Added public method to `Regexbuilder` called `case_insensitive` 
		- This will allow for forcing of a flag if not present in the raw str
- Added a call to the `RaRegexBuilder` which will consume this option using `.syntax( ... )`
- Added tests into `lib.rs` to make sure the override works as intended 

### Examples

Currently, we can write in flags into the raw string like this:
```rust 
use fancy_regex::RegexBuilder;

let re = r"(?i)FOO BAR";
let regex = RegexBuilder::new(re).build()?;
assert!(regex.is_match("foo bar").unwrap_or_default());
```
This change is to enable builder functionality like this:
```rust 
use fancy_regex::RegexBuilder;

let re = r"FOO BAR";
let regex = RegexBuilder::new(re)
	.case_insensitive(true)
	.build()?;
assert!(regex.is_match("foo bar").unwrap_or_default());
```

# Notes

Currently, `fancy-regex` doesn't support flags such as `(?-i)` and that is likely something we should enable in the future. The correlation of that type of flag to this feature add ties to the builder `case_insensitive` setter method. This method should override any raw input data.


src/compile.rs
```diff
@@ -499,6 +499,7 @@ pub(crate) fn compile_inner(inner_re: &str, options: &RegexOptions) -> Result<Ra
 
     let re = RaBuilder::new()
         .configure(config)
+        .syntax(options.syntaxc)
```
robinst, line 502: Is this the only place where we actually construct a case insensitive matcher at the moment?

jonperry-dev: To the best of my knowledge and search, I believe this is the only place that we need this.

jonperry-dev: @robinst Just wanted to bump this, was there anything you feel like is missing currently for `lowercase` specific building? 

I'd like to iterate over the rest of the builder functionality as a future works.

robinst: Hey @jonperry-dev, thanks for your patience. Yeah, let's go ahead with this. Can you rebase the PR? The builds seem to have expired, sorry about that.

robinst (approved):

src/compile.rs
jonperry-dev: No worries at all! Thanks for getting back to me. 

Just updated, should be good to go.

I hope everything is well! 🙂

jonperry-dev: @robinst I think its now waiting for maintainer approval on CI. I did a few more changes: 
- Cleaned up the code
- had a `cargo fmt` change
- moved tests to its own file (plan on making more extensions to the options)

jonperry-dev:
@robinst I'm not sure if that coverage failure with line 505 is an issue? I'm not entirely sure what its looking for there.

robinst:
> @robinst I'm not sure if that coverage failure with line 505 is an issue? I'm not entirely sure what its looking for there.

Seems like tarpaulin (the tool we use for coverage) doesn't analyze that correctly. (Maybe bumping the version would help, not sure.)

robinst:
Thanks, merged!

robinst:
Released in 0.14.0: https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md#0140---2024-10-24

PR #140, opened 2024-09-24
Add Split Struct and split Method to Regex

### LINKED ISSUE #104: split function missing
i wanted to split a string with regex and was wondering if theres a reason that that function is missing? its implemented in the rust-regex crate and also their regex implements pattern so the std::split is an option as well. 
i am interested in adding it but wondering if it was purposely left out for some reason?

k94-ishi:
Closes #104

This pull request introduces a Split struct and a corresponding split method to the Regex struct. This enhancement allows users to split a target string by the matches of a regex, yielding an iterator of substrings that are not matched by the regex.

(I referenced the code in #104 .)

## Changes

- **Created `Split` Struct**: Implemented the `Iterator` trait for the `Split` struct, allowing it to return the next substring from the target string based on regex matches.
- **Added `split` Method**: Implemented the `split` method in the `Regex` struct to return an instance of the `Split` iterator.
- **Documentation**: Provided documentation for the `split` method, including usage examples.
- **Test Cases**: Added test cases in `tests/splitting.rs` to verify the functionality of the `split` method.

## Examples
Here are some examples of how to use the `split` method:

```rust
// Example usage of the new split functionality
use fancy_regex::Regex;

let re = Regex::new(r"[ \t]+").unwrap();
let target = "a b \t  c\td    e";
let fields: Vec<&str> = re.split(target).map(|x| x.unwrap()).collect();
assert_eq!(fields, vec!["a", "b", "c", "d", "e"]);
```

## Note

- In this method, an empty string is taken out as an element as in the following example. This specification conforms to the behavior of str::split.

``` rust
#[test]
fn split_by_own() {
    let re = Regex::new("123").unwrap();
    let target = "123";
    let fields: Vec<&str> = re.split(target).map(|x| x.unwrap()).collect();
    assert_eq!(fields, vec!["", ""]);
}
```

- If there are no problems with this feature addition, it should be possible to implement the same for SplitN.
- I was particularly confused about deciding on variable names, is that ok?

### DIFF
--- a/tests/splitting.rs
+++ b/tests/splitting.rs
@@ -1,112 +1,93 @@
 use fancy_regex::Regex;
 
+fn split_to_vec<'a>(re_str: &'a str, target: &'a str) -> Vec<&'a str> {
+    let re = Regex::new(re_str).unwrap();
+    re.split(target).map(|x| x.unwrap()).collect()
+}
+
 #[test]
-fn split_left_center_right() {
-    let re = Regex::new("1").unwrap();
-    let target = "123";
-    let fields: Vec<&str> = re.split(target).map(|x| x.unwrap()).collect();
-    assert_eq!(fields, vec!["", "23"]);
+fn split_left() {
+    let result: Vec<&str> = split_to_vec("1", "123");
+    assert_eq!(result, vec!["", "23"]);
+}
 
-    let re = Regex::new("2").unwrap();
-    let fields: Vec<&str> = re.split(target).map(|x| x.unwrap()).collect();
-    assert_eq!(fields, vec!["1", "3"]);
+#[test]
+fn split_center() {
+    let result: Vec<&str> = split_to_vec("2", "123");
+    assert_eq!(result, vec!["1", "3"]);
+}
 
-    let re = Regex::new("3").unwrap();
-    let fields: Vec<&str> = re.split(target).map(|x| x.unwrap()).collect();
-    assert_eq!(fields, vec!["12", ""]);
+#[test]
+fn split_right() {
+    let result: Vec<&str> = split_to_vec("3", "123");
+    assert_eq!(result, vec!["12", ""]);
 }
 
 #[test]
 fn split_no_matches() {
-    let re = Regex::new("4").unwrap();
-    let target = "123";
-    let fields: Vec<&str> = re.split(target).map(|x| x.unwrap()).collect();
-    assert_eq!(fields, vec!["123"]);
+    let result: Vec<&str> = split_to_vec("4", "123");
+    assert_eq!(result, vec!["123"]);
 }
 
 #[test]
 fn split_empty() {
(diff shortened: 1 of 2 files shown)


k94-ishi:
I'm sorry for forgetting to run rust fmt before submitting the PR, even though it was noted in the CONTRIBUTING.md.

Also, it seems that the coverage test failed due to a rate limit error on Codecov: "Rate limit reached. Please upload with the Codecov repository upload token to resolve issue." Could you please rerun the coverage test when convenient?

keith-hall (approved):

k94-ishi:
@keith-hall 
Thanks for approving!
Just checking — is it okay to leave the Codecov upload error as is, or should I do something about it? Let me know if anything else is needed.

keith-hall:
I don't think you need to look into the codecov upload error. Hopefully our main maintainer will be able to look into it at some point. I'm just holding off merging to give @robinst chance to review. Thanks for your contribution btw! 🙂

k94-ishi:
Understood! Thanks for the quick reply! 🙂

tests/splitting.rs
```diff
@@ -0,0 +1,104 @@
+use fancy_regex::Regex;
+
+#[test]
+fn split_left_center_right() {
+    let re = Regex::new("1").unwrap();
+    let target = "123";
+    let fields: Vec<&str> = re.split(target).map(|x| x.unwrap()).collect();
+    assert_eq!(fields, vec!["", "23"]);
```
robinst, line 8: Can you add a helper function for this repeating pattern (if simple enough)?

tests/splitting.rs
```diff
@@ -0,0 +1,104 @@
...
+    let re = Regex::new(r"1(?=2)").unwrap();
+    let target = "12112122";
+    let fields: Vec<&str> = re.split(target).map(|x| x.unwrap()).collect();
+    assert_eq!(fields, vec!["", "21", "2", "22"]);
+
+    let re = Regex::new(r"(?<=2)1(?=2)").unwrap();
+    let target = "12112122";
+    let fields: Vec<&str> = re.split(target).map(|x| x.unwrap()).collect();
+    assert_eq!(fields, vec!["12112", "22"]);
+}
```
robinst, line 104: Can you also add a test for an empty regex?

robinst (approved):
Looks great, thank you! Just some minor inline comments, feel free to merge after addressing those.

I'll have a look at the Codecov error separately.

k94-ishi:
@robinst 
Thank you for the code review! 😃
I've addressed the suggested points, but I can't see the merge button in the GUI. It seems that the GitHub Actions workflow is awaiting approval. Could you please approve it and run the workflow?

k94-ishi:
Thanks for merging!
I’d like to implement splitn, but it’s not ready yet. If you could wait a little longer, I should be able to open a PR for it soon.
I appreciate your patience!

PR #148, opened 2024-12-07
Implement \Z

### LINKED ISSUE #147: Matching for \Z does not work
My regex: Regex::new(r"(?<=do\(\)|\A)[\s\S]*?(?=don\'t\(\)|\Z)").unwrap();

Error: thread 'main' panicked at src/main.rs:21:81:
called `Result::unwrap()` on an `Err` value: ParseError(36, InvalidEscape("\\Z"))

It seems like matching for end of string (\Z) does not work as expected.

CleanShot 2024-12-04 at 10 58 22@2x

keith-hall:
closes #147

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -357,8 +357,8 @@ impl<'a> Parser<'a> {
                 Expr::LookAround(
                     Box::new(Expr::Delegate {
                         inner: "\n*$".to_string(),
-                        size: 1,
-                        casei: self.flag(FLAG_CASEI),
+                        size: 0,
+                        casei: false,
                     }),
                     LookAhead,
                 ),
@@ -980,7 +980,7 @@ mod tests {
             Expr::LookAround(
                 Box::new(Expr::Delegate {
                     inner: "\n*$".to_string(),
-                    size: 1,
+                    size: 0,
                     casei: false,
                 }),
                 LookAhead,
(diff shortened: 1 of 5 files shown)


src/parse.rs
```diff
@@ -351,6 +351,18 @@ impl<'a> Parser<'a> {
...
         } else if b == b'z' && !in_class {
             (end, Expr::Assertion(Assertion::EndText))
+        } else if b == b'Z' && !in_class {
+            (
+                end,
+                Expr::LookAround(
+                    Box::new(Expr::Delegate {
+                        inner: "\n*$".to_string(),
+                        size: 1,
+                        casei: self.flag(FLAG_CASEI),
```
robinst, line 361: I think `size` should be 0, no? It's a zero-length match. (I don't recall how size for delegates is actually used, but yeah.)

Also, I think `casei` should just be `false` here, regardless of the flag. (`\n*$` never needs to be wrapped in `(?i:`.)

robinst (commented):
One comment, otherwise LGTM. Thanks!

src/parse.rs
keith-hall: Fixed. Also you were write about `size` - [it is used to determine `min_size`,](https://github.com/forkeith/fancy-regex/blob/c1b8a315239f07a2dc55dd1c5f0f22f64a5db03b/src/analyze.rs#L149) so 0 is indeed the correct size to use here.

Issue #141, opened 2024-09-29
Recursive Expressions (?R) in Rust's fancy-regex


shiueo:
I've looked through several regex crates in Rust (rust-onig / regex,,). but I haven't seen any that implement recursive expressions `(?R)` (to parse syntax like `{{{((?:[^{}]*|(?R))+)}}}`) yet. The README mentions that this feature is still in development, and while the basic ideas are in place, it states that procedure calls and recursive expressions are currently missing. 

Do you have any updates or timelines regarding when this functionality might be implemented?

keith-hall:
Hi, thanks for your interest. I have the idea that the readme is misleading unfortunately - I don't know of anyone actively working on this feature. And I personally don't think I will have time to work on recursive expressions any time soon. Contributions are very welcome, however 🙂

shiueo:
Thanks for the quick response and clarification!

keith-hall:
FWIW, [I have recently raised a PR with recursion support](https://github.com/fancy-regex/fancy-regex/pull/230), though the syntax is a little different than you asked about - instead of `(?R)` its more like `\g<0>` to recurse the 0th capture group aka whole expression.

PR #151, opened 2025-03-01
Update parser to support parsing subroutine call syntax

keith-hall:
I had the idea to implement subroutine calls, so this represents the first stage of the work - parsing the regex patterns.
Hopefully doing it in stages will make the PRs smaller and easier to review. Later will come the real changes to the analyzer, compiler and VM itself. This means I had to introduce a new `CompileError` to represent features which are recognized by the parser but are as yet unsupported for actually performing matches.

Partially inspired by:
- https://github.com/fancy-regex/fancy-regex/issues/141
- https://github.com/sharkdp/bat/issues/3156

Currently builds upon https://github.com/fancy-regex/fancy-regex/pull/148, as that improved test coverage etc.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -262,13 +262,7 @@ impl<'a> Parser<'a> {
                 },
             )),
             b'(' => self.parse_group(ix, depth),
-            b'\\' => {
-                let (next, expr) = self.parse_escape(ix, false)?;
-                if let Expr::Backref(group) = expr {
-                    self.backrefs.insert(group);
-                }
-                Ok((next, expr))
-            }
+            b'\\' => self.parse_escape(ix, false),
             b'+' | b'*' | b'?' | b'|' | b')' => Ok((ix, Expr::Empty)),
             b'[' => self.parse_class(ix),
             b => {
@@ -308,9 +302,7 @@ impl<'a> Parser<'a> {
                 None
             };
             if let Some(group) = group {
-                //if is_subroutine_call {
-                    self.backrefs.insert(group);
-                //}
+                self.backrefs.insert(group);
                 return Ok((ix + skip, if is_subroutine_call { Expr::SubroutineCall(group) } else { Expr::Backref(group) }));
             }
             // here the name is parsed but it is invalid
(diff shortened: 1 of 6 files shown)


src/parse.rs
```diff
@@ -307,7 +302,15 @@ impl<'a> Parser<'a> {
                 None
             };
             if let Some(group) = group {
-                return Ok((ix + skip, Expr::Backref(group)));
+                self.backrefs.insert(group);
```
robinst, line 305: I wasn't sure about this part initially (and `&self` -> `&mut self`) but thinking about alternatives, it seems best to modify backrefs here, at the place where we create the corresponding Expr, not in a higher place (depending on what we returned before, as the old code did).

src/parse.rs
```diff
@@ -307,7 +302,15 @@ impl<'a> Parser<'a> {
...
+                return Ok((
+                    ix + skip,
+                    if is_subroutine_call {
+                        Expr::SubroutineCall(group)
+                    } else {
+                        Expr::Backref(group)
+                    },
```
robinst, line 312: I'm not sure about this `boolean` `is_subroutine_call` argument though. Would this be nicer?:

```rust
    fn parse_named_backref<F>(
        &self,
        ix: usize,
        open: &str,
        close: &str,
        allow_relative: bool,
        create_expr: F,
    ) -> Result<(usize, Expr)> where F: FnOnce(usize) -> Expr {
        ...
    }
```

And then either:

```rust
return self.parse_named_backref(end, "'", "'", true, |group| Expr::Backref(group));
```

or:

```rust
self.parse_named_backref(end, "'", "'", true, |group| Expr::SubroutineCall(group))?
```

robinst (commented):
Nice. Just one suggestion.

Also, for stacked PRs (next time), you could have made this PR target the branch of the first PR instead of main. That way, the diff for this PR wouldn't have included both PRs.

src/parse.rs
keith-hall: thanks for the great suggestion! I agree, I much prefer this way of doing things, it does look nicer - updated :)

robinst (approved):
🎉

PR #152, opened 2025-03-10
Fix toy example

keith-hall:
The `toy` example had a number of shortcomings:
- the debug output was too verbose - it showed the PikeVM debug output instead of the regex pattern for example, making the graph unreadably big
- it panicked with regex patterns which include backrefs etc. due to the analyzed expression tree missing instructions for the 0th group

This PR fixes those problems and adds some tests which verify the output from the `toy` example, and runs those tests in CI.


robinst (approved):
Thank you!

PR #153, opened 2025-03-14
fix infinite loop when backtracking limit is hit

### LINKED ISSUE #137: `find_iter()` stuck in infinite loop after `Error::BacktrackLimitExceeded` occurs
The following code infinite loops.

```rust
use fancy_regex::RegexBuilder;

fn main() {
    let r = RegexBuilder::new("(x+x+)+(?=y)")
        .backtrack_limit(1)
        .build()
        .unwrap();
    let _: Vec<_> = r.find_iter("xxxxxxxxxxy").collect();
    println!("done!");
}
```

We encountered this in code that we didn't know was hitting the default backtracking limit. It's possible to catch this by changing the type of the collection to `Result<Vec<_>, _>`, but it seems like an easy mistake to make.

Given that it doesn't seem possible to recover from a `BacktrackLimitExceeded` error, would it be reasonable to have `find_iter()`and `captures_iter()` stop iteration when they encounter such an error? It could also simply be a documentation opportunity on `find_iter()` and `captures_iter()`.

keith-hall:
Fixes #137

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -302,7 +302,12 @@ impl<'r, 't> Iterator for Matches<'r, 't> {
                 .re
                 .find_from_pos_with_option_flags(self.text, self.last_end, option_flags)
             {
-                Err(error) => return Some(Err(error)),
+                Err(error) => {
+                    // Stop on first error: If an error is encountered, return it, and set the "last match position"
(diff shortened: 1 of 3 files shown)


robinst (approved):
Great work, thanks for fixing!

PR #154, opened 2025-04-04
Add support for named captures using (?'name'...) syntax

keith-hall:
We previously supported the angle bracket `(?<name>...)` syntax but missed the single quoted variant which Oniguruma supports

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -636,20 +636,12 @@ impl<'a> Parser<'a> {
         } else if self.re[ix..].starts_with("?<") || self.re[ix..].starts_with("?'") {
             // Named capture group using Oniguruma syntax: (?<name>...) or (?'name'...)
             self.curr_group += 1;
-            if let Some((id, skip)) = parse_id(
-                &self.re[ix + 1..],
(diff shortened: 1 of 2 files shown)


src/parse.rs
```diff
@@ -633,10 +633,23 @@ impl<'a> Parser<'a> {
...
+                if self.re[ix..].starts_with("?<") {
+                    "<"
+                } else {
+                    "'"
+                },
+                if self.re[ix..].starts_with("?<") {
+                    ">"
+                } else {
+                    "'"
+                },
```
robinst, line 650: Let's pull these out as variables, something like?:

```suggestion
let (open, close) = if self.re[ix..].starts_with("?<") {
                    ("<", ">")
                } else {
                    ("'", "'")
                };
```

robinst (commented):
Looks like codecov is broken again, will have a look separately.

src/parse.rs
keith-hall: great idea, thanks - done :slightly_smiling_face:

robinst (approved):

PR #155, opened 2025-04-04
Mark more expressions as non-repeatable

keith-hall:
to match Oniguruma, we should emit a parse error when quantifiers are used on assertions and other Expr nodes where it makes no sense to have a quantifier


robinst (approved):
Thank you!

PR #157, opened 2025-04-04
Update parser to support backrefs with a relative recursion level

keith-hall:
Basically a backref can either be to a relative group number like `(1)\k<-1>)` or it can be specifying a relative recursion level, but not both. When specifying a relative recursion level, the group number has to be absolute. So I updated `parse_id` to handle that, and callers then choose how to interpret the relative part based on the context.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -301,7 +301,14 @@ impl<'a> Parser<'a> {
             self.parse_named_backref_or_subroutine(ix, open, close, allow_relative)?;
         if let Some(group) = group {
             self.backrefs.insert(group);
-            return Ok((end, if let Some(recursion_level) = recursion_level { Expr::BackrefWithRelativeRecursionLevel(group, recursion_level) } else { Expr::Backref(group) }));
+            return Ok((
+                end,
+                if let Some(recursion_level) = recursion_level {
+                    Expr::BackrefWithRelativeRecursionLevel(group, recursion_level)
+                } else {
+                    Expr::Backref(group)
+                },
+            ));
         }
         if let Some(id) = id {
             // here the name was parsed but doesn't match a capture group we have already parsed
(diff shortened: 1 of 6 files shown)


src/parse.rs
```diff
@@ -985,41 +1003,57 @@ pub(crate) fn parse_decimal(s: &str, ix: usize) -> Option<(usize, usize)> {
...
+    if relative_sign == b'+' || relative_sign == b'-' {
+        if let Some((end, relative_amount)) = parse_decimal(&s, id_end + 1) {
+            if s[end..].starts_with(close) {
+                if relative_amount == 0 && id_len == 0 {
+                    return None;
+                }
+                let mut relative_amount_signed = relative_amount as isize;
+                if relative_sign == b'-' {
+                    relative_amount_signed = 0 - relative_amount_signed;
+                }
```
robinst, line 1047: Can we do without `mut`:

```suggestion
                let relative_amount_signed = if relative_sign == b'-' {
                    -(relative_amount as isize)
                } else {
                    relative_amount as isize
                };
                
```

src/parse.rs
```diff
@@ -985,41 +1003,57 @@ pub(crate) fn parse_decimal(s: &str, ix: usize) -> Option<(usize, usize)> {
         .map(|val| (end, val))
 }
 
-/// Attempts to parse an identifier between the specified opening and closing
-/// delimiters.  On success, returns `Some((id, skip))`, where `skip` is how much
-/// of the string was used.
+/// Attempts to parse an identifier, optionally followed by a relative number between the
+/// specified opening and closing delimiters.  On success, returns
+/// `Some((id, relative, skip))`, where `skip` is how much of the string was used.
```
robinst, line 1008: Here a struct would be nice too.

robinst (approved):
Just some suggestions, but feel free to address in separate PR.

src/parse.rs
```diff
@@ -339,32 +349,40 @@ impl<'a> Parser<'a> {
         unreachable!()
     }
 
+    // Returns Ok(ix, resolved group number, unresolved group name, recursion level)
```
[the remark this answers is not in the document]
keith-hall: totally agree :+1:

src/parse.rs
keith-hall: nice, yes we can :+1:

src/parse.rs
keith-hall: :+1:

PR #158, opened 2025-04-24
support \O to match any character including a newline

keith-hall:
this simple change gets a few more Oniguruma test cases passing, no need for any new Exprs or anything :)


robinst (approved):
Nice :)

PR #159, opened 2025-04-24
Improve toy compile output on wrapped regexes

keith-hall:
When a regex pattern is easy and fancy-regex is just used as a wrapper, the `toy compile` output was showing the debug output of the regex-automata compilation of the pattern instead of just the plain regex pattern string. This change fixes that, and makes it easier to understand the `toy compile` output in such situations.


keith-hall:
codecov isn't taking tests in examples into account it seems. I don't think it is worth duplicating some tests into other files just to make these code coverage failures go away, but let me know if you disagree...

robinst (approved):
LGTM. Yeah, ignore the coverage.

PR #160, opened 2025-04-24
Support case insensitive backrefs

keith-hall:


src/vm.rs
```diff
@@ -419,6 +424,15 @@ fn matches_literal(s: &str, ix: usize, end: usize, literal: &str) -> bool {
     end <= s.len() && &s.as_bytes()[ix..end] == literal.as_bytes()
 }
 
+#[inline]
+fn matches_literal_casei(s: &str, ix: usize, end: usize, literal: &str) -> bool {
+    end <= s.len()
+        && (matches_literal(s, ix, end, literal)
+            || (s.is_char_boundary(ix)
+                && s.is_char_boundary(end)
+                && s[ix..end].eq_ignore_ascii_case(literal)))
```
keith-hall, line 433: I'm not convinced this is the best way to achieve this, and likely would benefit from some tests where it would not be a utf8 char boundary, but it passes the Oniguruma test cases.

robinst: 1. Can you check how Oniguruma handles Unicode? E.g. for a regex like `(δ)(?i:\\1)`, does it match the text `δΔ` (I suspect yes)?
2. I think we should delegate to the regex crate's implementation of case-insensitive matching, to be consistent with other cases. The question is, how do we do that most efficiently given the literal we're matching against comes from a match during vm execution. I had a look through regex-automata and regex-syntax, and it looks like case folding happens in regex-syntax (I think [here](https://github.com/rust-lang/regex/blob/1a069b9232c607b34c4937122361aa075ef573fa/regex-syntax/src/hir/translate.rs#L391)). I wonder what the cheapest way of creating a regex that matches a literal is (i.e. if we could bypass parsing, that would be cool; instead of having to escape the literal and then go through full regex parsing before matching).

keith-hall: Regarding number 1, yep, you are right, Oniguruma finds a match

With number 2, I believe we can construct `Hir` directly, to avoid having to escape the literal and go through the parser again 👍

robinst: > With number 2, I believe we can construct `Hir` directly, to avoid having to escape the literal and go through the parser again 👍

Sounds good!

keith-hall: done :)

src/vm.rs
```diff
@@ -419,6 +424,60 @@ fn matches_literal(s: &str, ix: usize, end: usize, literal: &str) -> bool {
     end <= s.len() && &s.as_bytes()[ix..end] == literal.as_bytes()
 }
 
+#[inline]
```
robinst, line 427: I'd probably not force inlining of this as it's pretty long (and probably not used often). AFAIK leaving it off will mean that the compiler will decide.

src/vm.rs
```diff
@@ -419,6 +424,60 @@ fn matches_literal(s: &str, ix: usize, end: usize, literal: &str) -> bool {
...
+    }
+    if s.is_char_boundary(ix) && s.is_char_boundary(end) && s.is_ascii() {
+        return s[ix..end].eq_ignore_ascii_case(literal);
+    }
+
+    // text captured and being backreferenced is not ascii, so we utilize regex-automata's case insensitive matching
+
+    let chars = literal.chars();
+
+    use alloc::boxed::Box;
```
robinst, line 443: Is this `use` necessary?

src/vm.rs
```diff
@@ -419,6 +424,60 @@ fn matches_literal(s: &str, ix: usize, end: usize, literal: &str) -> bool {
...
+            }],
+        }),
+        ast: Box::new(Ast::Concat(Box::new(Concat {
+            span,
+            asts: literals,
+        }))),
+    }));
+
+    let mut translator = regex_syntax::hir::translate::Translator::new();
+    let hir = translator.translate(literal, &ast_flags).unwrap();
```
robinst, line 472: Looks like we can simplify this a bit, like this:

```rust
    let span = Span::splat(Position::new(0, 0, 0));
    let literals = literal
        .chars()
        .map(|c| {
            Ast::literal(Literal {
                span,
                kind: LiteralKind::Verbatim,
                c: c,
            })
        })
        .collect();
    let ast = Ast::concat(Concat {
        span,
        asts: literals,
    });

    let mut translator = regex_syntax::hir::translate::TranslatorBuilder::new()
        .case_insensitive(true)
        .build();
    let hir = translator.translate(literal, &ast).unwrap();
```

(Note that using `map` and `collect` means we don't need to give our own size hint to the vector, `Chars` does it for us.)

robinst (approved):
Some suggestions :)

src/vm.rs
keith-hall: good point, thanks, fixed :)

src/vm.rs
keith-hall: it was (for `--no-default-features`) but since I applied your suggestion it become unnecessary :tada:

src/vm.rs
keith-hall: thanks, I like! :rocket: done! :slightly_smiling_face:

src/vm.rs
robinst: Oh I see :)

PR #164, opened 2025-06-23
all extra flags to be set

hugopendlebury:
Enables extra flags to be set such as multi-line, white_space and dot_all


hugopendlebury:
Issue with PR

Issue #98, opened 2022-07-14
Add support for multi_line builder option


tgross35:
`regex` has support for `RegexBuilder::multi_line()` ([source here](https://docs.rs/regex/latest/regex/struct.RegexBuilder.html#method.multi_line)), would it be possible to add that here? 

I think the implementation can be done with a simple transformation if full PCRE is enabled - `^` to `(?<=^|\r\n|\n|\x0b|\f|\r|\x85)` and `$` to `(?=$|\r\n|\n|\x0b|\f|\r|\x85)`

robinst:
Hey! You can already get that behavior by starting your regex with `(?m)`. A `multi_line` method would be convenient to match regex's API though. I've added a "help wanted" label to this issue if anyone wants to contribute that work.

ash0080:
Interesting,  ?m seems not work, t will stop before the first line break even if I only use  r".*"

robinst:
Ah, sounds like you want [dot_matches_new_line](https://docs.rs/regex/latest/regex/struct.RegexBuilder.html#method.dot_matches_new_line) which is `(?s)`:

```
$ cargo run --example toy run '(?s)(.*)' 'hello\nworld'
captures: 0:[0..12] "hello\nworld" 1:[0..12] "hello\nworld"
```

robinst:
Fixed in 0.15.0: https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md#0150---2025-07-06

Issue #163, opened 2025-06-20
Issue with case-insensitive regex


angelorendina:
I was trying to use the regex defined here [here](https://colinhacks.com/essays/reasonable-email-regex) to validate emails. The regex is explained in the reference, and it seems to work natively on the browser as is.
The issue I am facing is that the same regex (wrapped in case-insensitive fancy-regex) fails if there are capitalised letters, see repro below.

```rust
fn is_valid_email(email: &str) -> bool {
    let regex = fancy_regex::RegexBuilder::new(
        r"^(?!\.)(?!.*\.\.)([a-z0-9_'+\-\.]*)[a-z0-9_'+\-]@([a-z0-9][a-z0-9\-]*\.)+[a-z]{2,}$",
    )
    .case_insensitive(true)
    .build()
    .unwrap();

    regex.is_match(email).unwrap()
}

#[test]
fn case_insensitive_email() {
    let is_valid = is_valid_email("VALID@domain.com");
    assert!(is_valid);
}
```

The test assertion fails (no panics), even though I am expecting it to pass (and the same regex with the same input are fine on my browser JS engine). Using `valid@domain.com` is perfectly accepted instead.

keith-hall:
Thanks for reporting, I can confirm that this is a bug. I prepared a test case (for https://github.com/fancy-regex/fancy-regex/blob/main/tests/regex_options.rs) to prove it, and it currently fails:
```rust
#[test]
fn check_casing_insensitive_option_hard() {
    let builder = RegexBuilder::new(r"[a-z](?<=[^f])")
        .case_insensitive(true)
        .build();

    match builder {
        Ok(regex) => {
            assert!(regex.is_match(r"J").unwrap_or_default());
            assert!(!regex.is_match(r"F").unwrap_or_default());
            assert!(regex.is_match(r"j").unwrap_or_default());
        }
        _ => panic!("builder should be able to compile with casing options"),
    }
}
```

The problem arises because `Compiler` always creates a new instance of `RegexOptions`:
https://github.com/fancy-regex/fancy-regex/blob/main/src/compile.rs#L105
https://github.com/fancy-regex/fancy-regex/blob/main/src/compile.rs#L527-L528

which therefore loses the case insensitive setting.
I had a quick look at fixing it by passing the RegexOptions as an argument to the ``compile` method, but it gets complicated because `compile` is `pub` so that the `toy`  example can use it, but `RegexOptions` isn't.
[...]

PR #165, opened 2025-06-23
Add multiline option and ignore space

hugopendlebury:
See
https://github.com/fancy-regex/fancy-regex/issues/98

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -557,11 +557,9 @@ impl RegexOptions {
         let whitespace =
             Self::get_flag_value(self.syntaxc.get_ignore_whitespace(), FLAG_IGNORE_SPACE);
         let dotnl = Self::get_flag_value(self.syntaxc.get_dot_matches_new_line(), FLAG_DOTNL);
-        let swap_greed = Self::get_flag_value(self.syntaxc.get_swap_greed(), FLAG_SWAP_GREED);
         let unicode = Self::get_flag_value(self.syntaxc.get_unicode(), FLAG_UNICODE);
 
-        let all_flags =
-            insensitive | multiline | whitespace | dotnl | swap_greed | unicode | unicode;
+        let all_flags = insensitive | multiline | whitespace | dotnl | unicode | unicode;
         all_flags
     }
 }
@@ -624,17 +622,6 @@ impl RegexBuilder {
     /// except for a new line character.
     pub fn dot_matches_new_line(&mut self, yes: bool) -> &mut Self {
         self.set_config(|x| x.dot_matches_new_line(yes))
-    }
-
-    /// Enable or disable the "swap greed" flag by default.
-    ///
-    /// When this is enabled, `.*` (for example) will become ungreedy and `.*?`
-    /// will become greedy.
-    ///
-    /// By default this is disabled. It may alternatively be selectively
-    /// enabled in the regular expression itself via the `U` flag.
-    pub fn swap_greed(&mut self, yes: bool) -> &mut Self {
-        self.set_config(|x| x.swap_greed(yes))
     }
 
     /// Enable verbose mode in the regular expression.
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -2681,46 +2681,4 @@ mod tests {
             ])
         );
     }
-
-    #[test]
-    fn parse_with_swap_greed_in_pattern() {
-        let tree = Expr::parse_tree("(?U)A.*Z");
-        let expr = tree.unwrap().expr;
-
-        assert_eq!(
-            expr,
-            Expr::Concat(vec![
-                make_literal("A"),
-                Expr::Repeat {
-                    child: Box::new(Expr::Any { newline: false }),
-                    lo: 0,
-                    hi: usize::MAX,
-                    greedy: false
-                },
-                make_literal("Z"),
-            ])
-        );
-    }
-
-    #[test]
-    fn parse_with_swap_greed_options() {
-        let options = get_options("A.*Z", |x| x.swap_greed(true));
-
-        let tree = Expr::parse_tree_with_flags(&options.pattern, options.compute_flags());
-        let expr = tree.unwrap().expr;
-
-        assert_eq!(
-            expr,
-            Expr::Concat(vec![
-                make_literal("A"),
-                Expr::Repeat {
-                    child: Box::new(Expr::Any { newline: false }),
-                    lo: 0,
-                    hi: usize::MAX,
-                    greedy: false
-                },
-                make_literal("Z"),
-            ])
-        );
-    }
 }
--- a/tests/regex_options.rs
+++ b/tests/regex_options.rs
@@ -89,6 +89,19 @@ fn check_ignore_whitespace_with_lookahead_matches() {
 
     assert!(!regex.is_match(test_text).unwrap_or_default());
 }
+
+#[test]
+fn check_verbose_mode_option() {
+    let pattern = "
+test  foo #hugo
+";
+    let regex = build_regex(RegexBuilder::new(pattern).verbose_mode(true));
+
+    let test_text = r"test    foo";
+
+    assert!(!regex.is_match(test_text).unwrap_or_default());
+}
+
 #[test]
 fn issue_163_fancy_email_test() {
     let pattern =


tests/regex_options.rs
```diff
@@ -42,3 +42,36 @@ fn check_casing_insensitive_option() {
...
+        .ignore_whitespace(true)
+        .build();
+
+    let test_text = r"test 
+
+    foo";
+
+    match builder {
+        Ok(regex) => assert!(regex.is_match(r"test foo").unwrap_or_default()),
+        _ => panic!("builder should be able to compile with casing options"),
```
keith-hall:
```suggestion
        _ => panic!("builder should be able to compile with ignore whitespace option"),
```

tests/regex_options.rs
```diff
@@ -42,3 +42,36 @@ fn check_casing_insensitive_option() {
...
+        .multi_line(true)
+        .build();
+
+    let test_text = r"test 
+
+    foo";
+
+    match builder {
+        Ok(regex) => assert!(regex.is_match(r"test foo").unwrap_or_default()),
+        _ => panic!("builder should be able to compile with casing options"),
```
keith-hall:
```suggestion
        _ => panic!("builder should be able to compile with multiline option"),
```

src/lib.rs
```diff
@@ -580,6 +580,30 @@ impl RegexBuilder {
...
+    pub fn ignore_whitespace(&mut self, yes:bool) -> &mut Self {
+        let syntaxc = self.0.syntaxc.to_owned();
+        self.0.syntaxc = syntaxc.ignore_whitespace(yes);
+        self
+    }
+
+    ///Enable or disable the "dot matches any character" flag by default.
+    ///When this is enabled, . will match any character. When it's disabled, then . will match any character 
+    /// except for a new line character.
+    pub fn dot_matches_new_line(&mut self, yes:bool) -> &mut Self {
```
keith-hall: I don't see a new unit test for this one, let's add it :)

tests/regex_options.rs
```diff
@@ -42,3 +42,36 @@ fn check_casing_insensitive_option() {
...
+    let builder = RegexBuilder::new(r"test.*foo")
+        .multi_line(true)
+        .build();
+
+    let test_text = r"test 
+
+    foo";
+
+    match builder {
+        Ok(regex) => assert!(regex.is_match(r"test foo").unwrap_or_default()),
```
keith-hall: am I missing something, or is `test_text`variable unused?

tests/regex_options.rs
```diff
@@ -42,3 +42,36 @@ fn check_casing_insensitive_option() {
...
+    let builder = RegexBuilder::new(r"  test.*foo")
+        .ignore_whitespace(true)
+        .build();
+
+    let test_text = r"test 
+
+    foo";
+
+    match builder {
+        Ok(regex) => assert!(regex.is_match(r"test foo").unwrap_or_default()),
```
keith-hall: I wonder if a clearer example (to show that whitespace in the regex pattern is ignored) would be:
```suggestion
    let builder = RegexBuilder::new(r"test    foo")
        .ignore_whitespace(true)
        .build();

    let test_text = r"testfoo";

    match builder {
        Ok(regex) => assert!(regex.is_match(test_text).unwrap_or_default()),
```

tests/regex_options.rs
hugopendlebury: whoops, good spot. Sadly when the text is corrected it panics. Not sure why.

src/lib.rs
hugopendlebury: added test

tests/regex_options.rs
hugopendlebury: now fixed... seems the expression was wrong and needs to have ^ and $

keith-hall:
Thanks for the improvements! I am still personally a little hesitant to merge this for the reasons mentioned in #163 - these options are not taken into account by the parser. So, `fancy-regex` parses the pattern (without knowing about ignore whitespace mode for example), then if it contains fancy features, builds a VM and delegates some parts of the matching to `regex-automata`. But this means that if users/consumers start relying on these new options, some regular expressions will be matched incorrectly.
Concrete example - this test panics when it shouldn't:
```rust
#[test]
fn check_ignore_whitespace_option_fancy() {

    let builder = RegexBuilder::new(r"(?=test    foo)")
        .ignore_whitespace(true)
        .build();

    let test_text = r"testfoo";
    match builder {
        Ok(regex) => assert!(regex.is_match(test_text).unwrap_or_default()),
        _ => panic!("builder should be able to compile with ignore whitespace option"),
    }
}
```

This test doesn't panic when it should:
```rust
#[test]
fn check_ignore_whitespace_option_fancy_wrong() {

    let builder = RegexBuilder::new(r"(?=test    foo)")
        .ignore_whitespace(true)
        .build();

    let test_text = r"test    foo";
    match builder {
        Ok(regex) => assert!(regex.is_match(test_text).unwrap_or_default()),
        _ => panic!("builder should be able to compile with ignore whitespace option"),
    }
}
```

hugopendlebury:
Ok I think I'm beginning to piece my head around what is going on.

WRT to 160 - Passing the RegexOptions only has partial success, it appears to work on the email but doesn't help with the test case you mention here.

I was wondering if a better approach might be to instead use the AST / Parsed Expression. 
With has nodes such as.

    Literal {
        /// The string to match
        val: String,
        /// Whether match is case-insensitive or not
        casei: bool,
    },

Would it be crazy to use to this AST / modify this tree ?

e.g. with this example

RegexBuilder::new(r"hello")
        .case_insensitive(true)
        .build();

Could the Tree be modified to consider any options ? Since the options would be set a global level ?

Perhaps this could be then be used to modify the regex. I would imagine that in case a regex is determined to be "fancy" the tree is passed to the automata library and for simple regex the passed expression could be modified to include the option
e.g. (?i)hello would be passed.

hugopendlebury:
> Wrapping nodes in `(?i)` should already be handled by the code which delegates to regex-automata. The main thing we need to achieve is, as you say, getting the AST correct. We should be able to do this by setting initial flags when parsing the regex pattern into an AST.

Ok I think I've got a clearer understanding of more or less what is needed. I will park this for now and see if I can look more at 163 and the AST. Since that appears to be more of a key for things working consistently.

keith-hall (approved):
This certainly solves the problem now, nice work, thanks for keeping at it. There are a few things that still make me hesitate a little, like making `RegexOptions` public and having the whole of `SyntaxConfig` passed to the parser when we only need certain flags. But, as it solves some very real issues, I would be tempted to say we could merge it and deal with those later, but let's see what @robinst thinks 🙂

hugopendlebury:
Thanks @keith-hall. 

Yeah I agree and commented as much. Let me see if there is some way to address that.

src/lib.rs
```diff
@@ -532,6 +535,7 @@ impl<'r, 'h> Iterator for SplitN<'r, 'h> {
 impl<'r, 'h> core::iter::FusedIterator for SplitN<'r, 'h> {}
 
 #[derive(Clone, Debug)]
+/// Allows the settings of Regex Options
```
keith-hall: Probably we can remove this doc comment again 🙂

src/lib.rs
```diff
@@ -540,6 +544,26 @@ struct RegexOptions {
...
+    fn get_flag_value(flag_value: bool, enum_value: u32) -> u32 {
+        if flag_value {
+            enum_value
+        } else {
+            0
+        }
+    }
+
+    fn compute_flags(&self) -> u32 {
+        let insenstive = Self::get_flag_value(self.syntaxc.get_case_insensitive(), FLAG_CASEI);
```
keith-hall: Small typo in the variable name 🙂

src/parse.rs
```diff
@@ -2484,4 +2501,63 @@ mod tests {
...
+        options
+    }
+
+    #[test]
+    fn parse_with_string_insenstive() {
+        let tree = Expr::parse_tree("(?i)hello");
+        let expr = tree.unwrap().expr;
+        let mut s = String::from("");
+        let s_ref: &mut String = &mut s;
+        expr.to_str(s_ref, 0);
```
keith-hall: Maybe we shouldn't test `to_str` but the Expr itself, not sure. Otherwise if we change the implementation, the test could fail as it is tightly coupled

src/lib.rs
hugopendlebury: corrected

src/parse.rs
hugopendlebury: Amended and added a few combined tests using multiple flags and combinations with pattern based options and RegexOptions / Config based settings.

src/parse.rs
```diff
@@ -1134,9 +1151,13 @@ fn is_hex_digit(b: u8) -> bool {
 }
 
 pub(crate) fn make_literal(s: &str) -> Expr {
+    make_literal_with_case_senitivity(s, false)
+}
+
+pub(crate) fn make_literal_with_case_senitivity(s: &str, case_insensitive: bool) -> Expr {
```
keith-hall: Small typo
```suggestion
pub(crate) fn make_literal_with_case_sensitivity(s: &str, case_insensitive: bool) -> Expr {
```

tests/regex_options.rs
```diff
@@ -42,3 +42,102 @@ fn check_casing_insensitive_option() {
...
+    let test_text = r"testfoo";
+
+    match builder {
+        Ok(regex) => assert!(regex.is_match(test_text).unwrap_or_default()),
+        _ => panic!("builder should be able to compile with ignore whitespace option"),
+    }
+}
+
+#[test]
+fn check_ignore_whitespace_option_fancy_wrong() {
```
keith-hall: Can we think of a better name for this test please? 🙂

keith-hall: nice, thanks!

src/lib.rs
```diff
@@ -569,15 +593,36 @@ impl RegexBuilder {
...
+    pub fn multi_line(&mut self, yes: bool) -> &mut Self {
+        self.set_config(|x| x.multi_line(yes))
+    }
+
+    /// Allow ignore whitespace
+    pub fn ignore_whitespace(&mut self, yes: bool) -> &mut Self {
+        self.set_config(|x| x.ignore_whitespace(yes))
+    }
+
+    ///Enable or disable the "dot matches any character" flag by default.
```
keith-hall: In other doc comments we have a space between the `///`and the text. Let's try to keep it consistent :)

src/lib.rs
```diff
@@ -569,15 +593,36 @@ impl RegexBuilder {
...
+    ///When this is enabled, . will match any character. When it's disabled, then . will match any character
+    /// except for a new line character.
```
keith-hall: as Markdown is supported in doc comments, probably it would look better on docs.rs if we wrap `.` in backticks.

src/lib.rs
hugopendlebury: done

src/lib.rs
hugopendlebury: done

src/lib.rs
```diff
@@ -203,12 +204,14 @@ mod compile;
 mod error;
 mod expand;
 mod parse;
+mod regexflags;
```
robinst: Maybe just `flags`?

src/lib.rs
```diff
@@ -188,6 +188,7 @@ use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
 
+use core::cmp::min;
```
robinst: Unused import?

src/lib.rs
```diff
@@ -569,15 +593,36 @@ impl RegexBuilder {
         Regex::new_options(self.0.clone())
     }
 
+    fn set_config(&mut self, func: impl Fn(SyntaxConfig) -> SyntaxConfig) -> &mut Self {
+        let syntaxc = self.0.syntaxc.to_owned();
+        self.0.syntaxc = func(syntaxc);
```
robinst: I think you just moved the code, but I don't think `to_owned` is necessary?:

```suggestion
        self.0.syntaxc = func(self.0.syntaxc);
```

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fancy-regex"
 # remember to update html_root_url
-version = "0.14.0"
+version = "0.14.1"
```
robinst: Let's not bump yet, we'll do a release separately (also need to update CHANGELOG, etc). And I think we should do 0.15.0 instead, as we're adding new features.

src/parse.rs
```diff
@@ -97,6 +90,14 @@ impl<'a> Parser<'a> {
...
 
+    /// Parse the regex and return an expression (AST) and a bit set with the indexes of groups
+    /// that are referenced by backrefs.
+    pub(crate) fn parse(re: &str) -> Result<ExprTree> {
+        let mut options = RegexOptions::default();
+        options.pattern = String::from(re);
+        Self::parse_with_flags(&options.pattern, options.compute_flags())
+    }
+
     fn new(re: &str) -> Parser<'_> {
```
robinst: This is not used anymore, right? Remove please. Then `new_with_flags` can just become `new`.

src/parse.rs
```diff
@@ -97,6 +90,14 @@ impl<'a> Parser<'a> {
         })
     }
 
+    /// Parse the regex and return an expression (AST) and a bit set with the indexes of groups
+    /// that are referenced by backrefs.
+    pub(crate) fn parse(re: &str) -> Result<ExprTree> {
+        let mut options = RegexOptions::default();
+        options.pattern = String::from(re);
+        Self::parse_with_flags(&options.pattern, options.compute_flags())
```
robinst:
```suggestion
        Self::parse_with_flags(re, RegexOptions::default().compute_flags())
```

Also, we can just inline this in the one place this is called from, `Expr::parse_tree`. And so this frees up the plain `parse` name, so let's just use that for `parse(re: &str, flags: u32)`.

src/parse.rs
```diff
@@ -1134,9 +1151,13 @@ fn is_hex_digit(b: u8) -> bool {
 }
 
 pub(crate) fn make_literal(s: &str) -> Expr {
+    make_literal_with_case_sensitivity(s, false)
+}
+
+pub(crate) fn make_literal_with_case_sensitivity(s: &str, case_insensitive: bool) -> Expr {
```
robinst: The name `make_literal_with_case_sensitivity` here is the wrong way around, e.g. `make_literal_with_case_sensitivity(true)` does the opposite of what it sounds like. Rename to `make_literal_case_insensitive`?

src/parse.rs
```diff
@@ -2484,4 +2505,197 @@ mod tests {
...
+
+    fn get_options(pattern: &str, func: impl Fn(SyntaxConfig) -> SyntaxConfig) -> RegexOptions {
+        let mut options = RegexOptions::default();
+        options.syntaxc = func(options.syntaxc);
+        options.pattern = String::from(pattern);
+        options
+    }
+
+    #[test]
+    fn parse_with_string_insenstive() {
```
robinst: Typos in this and next test names. Let's use this pattern for names:

- `parse_with_case_insensitive_in_pattern`
- `parse_with_case_insensitive_option`
- `parse_with_multiline_in_pattern`
- `parse_with_multiline_option`
- etc

tests/regex_options.rs
```diff
@@ -42,3 +42,102 @@ fn check_casing_insensitive_option() {
...
+    let builder = RegexBuilder::new(r"^test$").multi_line(true).build();
+
+    let test_text = r"test
+hugo
+test";
+
+    match builder {
+        Ok(regex) => assert!(regex.is_match(test_text).unwrap_or_default()),
+        _ => panic!("builder should be able to compile with multiline option"),
+    }
```
robinst: I can see this follows the existing pattern, but it would be nice to have some helper here to assert that building the regex succeeds. There is already a `regex` helper in `common/mod.rs`, can we add another one:

```rust
pub fn build_regex(builder: &RegexBuilder) -> Regex {
    let result = builder.build();
    assert!(
        result.is_ok(),
        "Expected regex to build successfully, got {:?}",
        result.err()
    );
    result.unwrap()
}
```

Then we can condense this test to:

```rust
let regex = build_regex(RegexBuilder::new(r"^test$").multi_line(true));
assert!(regex.is_match(test_text).unwrap_or_default());
```

robinst (changes requested):
Thanks for working on this! I have requested some changes. LGTM after those are addressed.

src/lib.rs
hugopendlebury: renamed

src/lib.rs
hugopendlebury: removed

Cargo.toml
hugopendlebury: reverted

src/parse.rs
hugopendlebury: renamed.

src/parse.rs
hugopendlebury: done

tests/regex_options.rs
hugopendlebury: done

src/parse.rs
hugopendlebury: done

robinst (approved):
LGTM now, thanks!

robinst:
@keith-hall do you want to prepare a PR for doing the next release? I think you have the most changes to be documented in the CHANGELOG :).

PR #167, opened 2025-07-07
optimize positive lookahead at end of pattern

keith-hall:
Introducing `fancy-regex`'s first optimization! This is done purely on the parse tree, without touching the VM.

We move a trailing positive lookahead outside our wrapped capture group 0.

If the contents of the lookahead are easy, and the rest of the expression is also easy, we can then delegate the whole regexp to regex-automata and fixup the capture groups afterwards.

If the contents are not easy, no harm is done because the expression retains the same semantics in our VM, as we only perform the move if the whole pattern is not self-recursive. Also, as it is done after parsing, any relative backreferences etc. should be unaffected.

Because it is done before analysis, no re-analysis needs to take place, so it is quite efficient.

This allows us to take full advantage of regex-automata's prefilters etc, and in general delegating to the NFA should always be quicker than using a backtracking engine and trying each character position in the haystack one at a time.

This type of regex pattern is a quite common scenario inside `.sublime-syntax` definitions which are compatible with Sublime Text's `sregex` engine.

Examples - the new capture group becomes our explicit capture group 0:
- `(?=foo)` becomes `()foo`
- `\w+(?=[ ])` becomes `(\w+)[ ]`
- `a(b)c(?=def)` becomes `(a(b)c)def`
- `a(b)c(?=def|ghi)` becomes `(a(b)c)(?:def|ghi)`

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,1 @@
-github: [raphlinus, robinst]
+github: [raphlinus, robinst, keith-hall]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html),
 with the exception that 0.x versions can break between minor versions.
 
+## [Unreleased]
+### Added
+- Add an optimization step after the pattern is parsed but before it is analyzed.
+  Currently it only optimizes one specific use-case - where the expression is easy except
+  for a trailing positive lookahead whose contents are also easy. The optimization is to avoid the VM.
+
 ## [0.15.0] - 2025-07-06
 ### Added
 - Support `\Z` - anchor to the end of the text before any trailing newlines. (#148)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ with the exception that 0.x versions can break between minor versions.
 ### Added
 - Add an optimization step after the pattern is parsed but before it is analyzed.
   Currently it only optimizes one specific use-case - where the expression is easy except
-  for a trailing positive lookahead whose contents are also easy. The optimization is to avoid the VM.
+  for a trailing positive lookahead whose contents are also easy. The optimization is to delegate to the regex crate instead of using the backtracking VM.
 
 ## [0.15.0] - 2025-07-06
 ### Added
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -240,10 +240,7 @@ enum RegexImpl {
     Wrap {
         inner: RaRegex,
         options: RegexOptions,
-    },
-    WrapCaptureFixup {
-        inner: RaRegex,
-        options: RegexOptions,
+        explicit_capture_group_0: bool,
     },
     Fancy {
         prog: Prog,
@@ -414,10 +411,7 @@ enum CapturesImpl<'t> {
     Wrap {
         text: &'t str,
         locations: RaCaptures,
-    },
-    WrapCaptureFixup {
-        text: &'t str,
-        locations: RaCaptures,
+        explicit_capture_group_0: bool,
     },
     Fancy {
         text: &'t str,
@@ -765,10 +759,10 @@ impl Regex {
             };
             let inner = compile::compile_inner(&re_cooked, &options)?;
             return Ok(Regex {
-                inner: if requires_capture_group_fixup {
-                    RegexImpl::WrapCaptureFixup { inner, options }
-                } else {
-                    RegexImpl::Wrap { inner, options }
+                inner: RegexImpl::Wrap {
+                    inner,
+                    options,
+                    explicit_capture_group_0: requires_capture_group_fixup,
                 },
                 named_groups: Arc::new(tree.named_groups),
             });
@@ -789,7 +783,6 @@ impl Regex {
     pub fn as_str(&self) -> &str {
         match &self.inner {
             RegexImpl::Wrap { options, .. } => &options.pattern,
-            RegexImpl::WrapCaptureFixup { options, .. } => &options.pattern,
             RegexImpl::Fancy { options, .. } => &options.pattern,
         }
     }
@@ -809,7 +802,6 @@ impl Regex {
     pub fn is_match(&self, text: &str) -> Result<bool> {
         match &self.inner {
             RegexImpl::Wrap { ref inner, .. } => Ok(inner.is_match(text)),
-            RegexImpl::WrapCaptureFixup { ref inner, .. } => Ok(inner.is_match(text)),
             RegexImpl::Fancy {
                 ref prog, options, ..
             } => {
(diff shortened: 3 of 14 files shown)


.github/FUNDING.yml
```diff
@@ -1 +1 @@
-github: [raphlinus, robinst]
+github: [raphlinus, robinst, keith-hall]
```
robinst, line 1: 🎉

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html),
 with the exception that 0.x versions can break between minor versions.
 
+## [Unreleased]
+### Added
+- Add an optimization step after the pattern is parsed but before it is analyzed.
+  Currently it only optimizes one specific use-case - where the expression is easy except
+  for a trailing positive lookahead whose contents are also easy. The optimization is to avoid the VM.
```
robinst:
```suggestion
  for a trailing positive lookahead whose contents are also easy. The optimization is to delegate to the regex crate instead of using the backtracking VM.
```

src/lib.rs
```diff
@@ -239,6 +241,10 @@ enum RegexImpl {
         inner: RaRegex,
         options: RegexOptions,
     },
+    WrapCaptureFixup {
```
robinst: Can you add some documentation here?

src/lib.rs
```diff
@@ -409,6 +415,10 @@ enum CapturesImpl<'t> {
         text: &'t str,
         locations: RaCaptures,
     },
+    WrapCaptureFixup {
```
robinst: Docs here as well, i.e. something like "capture group 1 in the API maps to group 2 internally, etc"

src/lib.rs
```diff
@@ -726,30 +736,40 @@ impl Regex {
...
+                let raw_e = match tree.expr {
+                    Expr::Concat(ref v) => match v[1] {
+                        Expr::Group(ref child) => child,
+                        _ => unreachable!(),
+                    },
                     _ => unreachable!(),
-                },
-                _ => unreachable!(),
+                };
+                raw_e.to_str(&mut re_cooked, 0);
```
robinst, line 760: I never liked this wrapping and unwrapping here, and this makes it a bit more complicated.

I wonder how complicated it would be to instead make the VM understand arbitrary start positions and capturing match bounds.. Do you know how the regex crate does this?

src/lib.rs
```diff
@@ -726,30 +736,40 @@ impl Regex {
...
+            } else {
+                tree.expr.to_str(&mut re_cooked, 0);
             };
-            raw_e.to_str(&mut re_cooked, 0);
             let inner = compile::compile_inner(&re_cooked, &options)?;
             return Ok(Regex {
-                inner: RegexImpl::Wrap { inner, options },
+                inner: if requires_capture_group_fixup {
+                    RegexImpl::WrapCaptureFixup { inner, options }
```
robinst: Instead of this we could also add a boolean to `Wrap`, right? Some code below could be simplified with that I think, e.g. in captures_from_pos.

src/lib.rs
```diff
@@ -1999,6 +2053,36 @@ mod tests {
...
     }
 
+    #[test]
+    fn trailing_positive_lookahead_wrap_capture_group_fixup() {
+        let s = r"(a+)(?=c)";
+        let regex = s.parse::<Regex>().unwrap();
+        match regex.inner {
+            RegexImpl::WrapCaptureFixup { .. } => {},
+            _ => panic!("trailing positive lookahead for an otherwise easy pattern should avoid going through the VM"),
+        }
```
robinst: Isn't there something to make this simpler, maybe `assert!(matches!(...))` (but keep the nice assertion message)?

src/lib.rs
```diff
@@ -726,30 +736,40 @@ impl Regex {
...
+                        Expr::Group(ref child) => child,
+                        _ => unreachable!(),
+                    },
                     _ => unreachable!(),
-                },
-                _ => unreachable!(),
+                };
+                raw_e.to_str(&mut re_cooked, 0);
+            } else {
+                tree.expr.to_str(&mut re_cooked, 0);
```
robinst: So in the case where we optimized, we're now asking the regex crate to match on`(?s:.)*?(pattern)`, instead of just `pattern`.. Does that change anything about how the matching works?

Basically I'm wondering whether the `(?s:.)*?` in the beginning has any bad effects (and we don't really need it for delegating, right?)

src/optimize.rs
```diff
@@ -0,0 +1,141 @@
...
+    if !tree.self_recursive {
+        let mut optimized_tree = tree.clone();
+        let requires_capture_group_fixup = optimize_trailing_lookahead(&mut optimized_tree)?;
+        Ok((optimized_tree, requires_capture_group_fixup))
+    } else {
+        Ok((tree, false))
+    }
+}
+
+fn optimize_trailing_lookahead(tree: &mut ExprTree) -> Result<bool> {
```
robinst: Why the `Result`?

src/optimize.rs
```diff
@@ -0,0 +1,141 @@
...
+    // returns a boolean to say whether the optimization was applied.
+    // - if it was applied, capture group 0 is no longer implicit, but explicit
+    //   if/when the whole expression gets delegated to regex-automata
+    // converts i.e. original pattern `a(?=b)` when wrapped in the capture group 0
+    // as `(?s:.)*?(a(?=b))`
```
robinst, line 45: Is `(?s:.)*?(a(?=b|c))` handled correctly here? I.e. does the content of the lookaround need to be wrapped in a group?

src/optimize.rs
keith-hall: good question, nice catch. No need for it - removed :)

src/optimize.rs
keith-hall: great question!  the `to_str` method handles that for us, I added some tests to prove it. (We actually don't have an `Expr` for a non-capture group)

src/lib.rs
keith-hall: I don't know, I haven't checked yet. Would you mind if we left it for a future PR if it is non-trivial?

src/lib.rs
keith-hall: yes, there is, thanks for the suggestion - done :)

src/lib.rs
keith-hall: done

src/lib.rs
keith-hall: please let me know if what I added is clear enough

src/lib.rs
keith-hall: done

src/lib.rs
keith-hall: I guess a bad effect could be extra memory usage on a large input haystack. We indeed don't need it for delegating... I've pushed a fix, but I'm not 100% sure I'm happy with it... Please see what you think...

src/lib.rs
keith-hall: the fix for https://github.com/fancy-regex/fancy-regex/pull/167#discussion_r2193379410 has perhaps made it even more complicated... so I will understand if you say "no" :sweat_smile:

robinst: The regex crate code for this is e.g. here:

https://github.com/rust-lang/regex/blob/1a069b9232c607b34c4937122361aa075ef573fa/regex-automata/src/nfa/thompson/pikevm.rs#L1218-L1219

Note that it knows whether a pattern is anchored or not, something we could do as well (another optimization).

Apart from that, I haven't gone back to this yet. I do now think it would not be too difficult to change the VM to get rid of the wrapping. The added benefit as mentioned above is that anchored matches that happen to not match would also be optimized.

(An alternative that I typed at some point but not sure I sent, is if we delayed the wrapping to later when we know we need the VM. AFAICT it would mean we'd have to wrap using `Infos` instead when we call `compile(&info)?`, but the nice thing is that the delegation case wouldn't be affected. Let me know if that makes sense or not 😅.)

keith-hall:
merged as part of https://github.com/fancy-regex/fancy-regex/pull/171

PR #170, opened 2025-07-20
Refactor to avoid wrapping exprtree

keith-hall:
as discussed in #167, this PR removes the need to wrap the `ExprTree` to get the match bounds or allow the match to start at any position. This opens up the optimization to perform an anchored search when the regex pattern begins with `^` meaning start of text (as opposed to start of line), by not emitting instructions to bump the index in the haystack. The end result being that it should fail faster for non-matches. (I'm leaving it as a future optimization to let regex-automata perform unanchored regex searches for the first `Delegate` etc.)

- instead of prepending `(?m:.*?)` to the `ExprTree`, add the instructions at compile time

Other changes:
- also ensure that benchmarks compile as part of CI
- the CompileError for an invalid (numbered) backref has been updated to mention which backref was invalid
- fixed a bug whereby sometimes a capture group containing a backref to itself would cause a compile error, when it is valid - this fixes a few Oniguruma test cases

### DIFF
--- a/src/analyze.rs
+++ b/src/analyze.rs
@@ -250,10 +250,25 @@ pub fn analyze<'a>(tree: &'a ExprTree, start_group: usize) -> Result<Info<'a>> {
     analyzed
 }
 
+/// Determine if the expression will always only ever match at position 0
+pub fn can_compile_as_anchored(root_expr: &Expr) -> bool {
+    use crate::Assertion;
+
+    match root_expr {
+        Expr::Concat(ref children) => match children[0] {
+            Expr::Assertion(ref assertion) => *assertion == Assertion::StartText,
+            _ => false,
+        },
+        Expr::Assertion(ref assertion) => *assertion == Assertion::StartText,
+        _ => false,
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::analyze;
     // use super::literal_const_size;
+    use crate::can_compile_as_anchored;
     use crate::CompileError;
     use crate::Error;
     use crate::Expr;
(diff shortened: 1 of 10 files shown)


src/analyze.rs
```diff
@@ -238,19 +235,40 @@ fn literal_const_size(_: &str, _: bool) -> bool {
...
+    if analyzer.backrefs.len() > analyzer.group_ix {
+        return Err(Error::CompileError(CompileError::InvalidBackref(
+            analyzer.backrefs.len() - 1,
+        )));
+    }
+    analyzed
+}
+
+/// Determine if the expression will always only ever match at position 0
+pub fn can_compile_as_anchored(root_expr: &Expr) -> bool {
```
robinst, line 256: Maybe add a note that this can return `false` even if it could be anchored. Basically `true` should only be treated as an optimization.

I'm thinking of cases like `^abc|^abb` or something like that :).

src/lib.rs
```diff
@@ -1861,28 +1849,11 @@ pub fn detect_possible_backref(re: &str) -> bool {
 }
 */
 
-/// wrapper to search for re at arbitrary start position,
-/// and to capture the match bounds
-pub fn wrap_tree(raw_tree: ExprTree) -> ExprTree {
```
robinst, line 1866: Nice :)

src/compile.rs
```diff
@@ -524,9 +524,22 @@ pub(crate) fn compile_inner(inner_re: &str, options: &RegexOptions) -> Result<Ra
 }
 
 /// Compile the analyzed expressions into a program.
-pub fn compile(info: &Info<'_>) -> Result<Prog> {
+pub fn compile(info: &Info<'_>, anchored: bool) -> Result<Prog> {
     let mut c = Compiler::new(info.end_group);
+    if !anchored {
+        c.b.add(Insn::Split(3, 1));
```
robinst: Maybe add some comments about these magic numbers, i.e. 3 means after these added instructions, 1 is the Any.

robinst (approved):
Great work! Just some suggestions

src/analyze.rs
keith-hall: good point, thanks - done :slightly_smiling_face:

src/compile.rs
keith-hall: good idea, it is much clearer what is happening that way - done

PR #171, opened 2025-07-26
Trailing positive lookahead optimization including wrap elimination

keith-hall:
Basically combines https://github.com/fancy-regex/fancy-regex/pull/167 and https://github.com/fancy-regex/fancy-regex/pull/170 (and adapts the trailing lookahead optimization now that the `ExprTree` isn't wrapped) - as I had made those PRs from my fork, it wasn't trivial to stack them...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -752,10 +752,7 @@ impl Regex {
                     inner,
                     options: RegexOptions {
                         pattern: re_cooked.clone(),
-                        syntaxc: options.syntaxc,
-                        backtrack_limit: options.backtrack_limit,
-                        delegate_size_limit: options.delegate_size_limit,
-                        delegate_dfa_size_limit: options.delegate_dfa_size_limit,
+                        ..options
                     },
                     explicit_capture_group_0: requires_capture_group_fixup,
                 },
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -733,10 +733,10 @@ impl Regex {
     }
 
     fn new_options(options: RegexOptions) -> Result<Regex> {
-        let tree = Expr::parse_tree_with_flags(&options.pattern, options.compute_flags())?;
+        let mut tree = Expr::parse_tree_with_flags(&options.pattern, options.compute_flags())?;
 
         // try to optimize the expression tree
-        let (tree, requires_capture_group_fixup) = optimize(tree);
+        let requires_capture_group_fixup = optimize(&mut tree);
         let info = analyze(&tree, if requires_capture_group_fixup { 0 } else { 1 })?;
 
         if !info.hard {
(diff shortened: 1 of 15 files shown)


src/lib.rs
```diff
@@ -727,17 +735,30 @@ impl Regex {
...
             return Ok(Regex {
-                inner: RegexImpl::Wrap { inner, options },
+                inner: RegexImpl::Wrap {
+                    inner,
+                    options: RegexOptions {
+                        pattern: re_cooked.clone(),
+                        syntaxc: options.syntaxc,
+                        backtrack_limit: options.backtrack_limit,
+                        delegate_size_limit: options.delegate_size_limit,
+                        delegate_dfa_size_limit: options.delegate_dfa_size_limit,
```
robinst: Can we use `..options` here instead?

src/optimize.rs
```diff
@@ -0,0 +1,180 @@
...
+
+use alloc::boxed::Box;
+use alloc::vec;
+use core::mem;
+
+/// Rewrite the expression tree to help the VM compile an efficient program.
+pub fn optimize(tree: ExprTree) -> (ExprTree, bool) {
+    // self recursion prevents us from moving the trailing lookahead out of group 0
+    if !tree.self_recursive {
+        let mut optimized_tree = tree.clone();
```
robinst: Hmm, why do we need the clone here? I think if we make `tree` a `mut` we can avoid the clone?

src/optimize.rs
```diff
@@ -0,0 +1,180 @@
...
+    // - if it was applied, capture group 0 is no longer implicit, but explicit
+    //   if/when the whole expression gets delegated to regex-automata
+    // converts i.e. original pattern `a(?=b)` when wrapped in the capture group 0
+    // as `(a(?=b))`
+    // to `(a)b`
+
+    if let Expr::Concat(ref mut root_concat_children) = tree.expr {
+        if let Some(last_child_of_root_concat) = &mut root_concat_children.last_mut() {
+            // we get the last child if it is a positive lookahead
+            if let Expr::LookAround(_, LookAround::LookAhead) = &last_child_of_root_concat {
```
robinst: I think we can combine this with the above, like this?:

```rust
if let Some(Expr::LookAround(_, LookAround::LookAhead)) = root_concat_children.last() {
    ...
}
```

robinst (approved):
LGTM, just some minor suggestions.

src/lib.rs
keith-hall: we certainly can, thanks :+1:

src/optimize.rs
keith-hall: done :+1:

src/optimize.rs
keith-hall: nice, we lose a level of nesting, thanks :slightly_smiling_face:

PR #174, opened 2025-08-01
Fix to ensure it always picks up invalid backrefs

### LINKED ISSUE #173: “Index out of bounds” caused by incorrect backreferences
Using **fancy-regex 0.16.0**, the next code detects an invalid backreference:

```rust
use fancy_regex::Regex;

fn main()  
{
    let re = Regex::new(r"a\1").unwrap();
    let result : bool = re.is_match("aa").unwrap();

    println!("{:?}", result);
}
```

It throws the expected error: _called \`Result::unwrap()\` on an \`Err\` value: ParseError(2, InvalidBackref)_.

However, if the pattern is changed:

```rust
let re = Regex::new(r"aa\1").unwrap();
let result : bool = re.is_match("aa").unwrap();
```

then it raises a different error: _thread 'main' panicked at C:\\Users\\...\\.cargo\\registry\\src\\index.crates.io-1949cf8c6b5b557f\\fancy-regex-0.16.0\\src\\vm.rs:348:19: index out of bounds: the len is 2 but the index is 2_.

I think that it should be generating the same _InvalidBackref_ error instead of “index out of bounds”.

In addition, if the code is changed to:

```rust
let re = Regex::new(r"aa\1").unwrap();
let result : bool = re.is_match("ab").unwrap();
```

then it does not report any problem, even though the backreference is still invalid. (The program prints _false_).

The program gives the expected _InvalidBackref_ error in all the above examples in case of previous **fancy-regex 0.15.0**, or if `\2` is used instead of `\1`.

keith-hall:
Fixes #173

### DIFF
--- a/src/analyze.rs
+++ b/src/analyze.rs
@@ -242,10 +242,14 @@ pub fn analyze<'a>(tree: &'a ExprTree, start_group: usize) -> Result<Info<'a>> {
     };
 
     let analyzed = analyzer.visit(&tree.expr);
-    if analyzer.backrefs.len() > analyzer.group_ix {
-        return Err(Error::CompileError(CompileError::InvalidBackref(
-            analyzer.backrefs.len() - 1,
-        )));
+    if let Some(highest_backref) = analyzer.backrefs.into_iter().last() {
+        if highest_backref > analyzer.group_ix - start_group
+            || highest_backref == analyzer.group_ix && start_group == 0
+        {
+            return Err(Error::CompileError(CompileError::InvalidBackref(
+                highest_backref,
+            )));
+        }
     }
     analyzed
 }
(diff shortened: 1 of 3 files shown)


src/analyze.rs
```diff
@@ -242,10 +242,14 @@ pub fn analyze<'a>(tree: &'a ExprTree, start_group: usize) -> Result<Info<'a>> {
     };
 
     let analyzed = analyzer.visit(&tree.expr);
-    if analyzer.backrefs.len() > analyzer.group_ix {
-        return Err(Error::CompileError(CompileError::InvalidBackref(
-            analyzer.backrefs.len() - 1,
-        )));
+    if let Some(highest_backref) = analyzer.backrefs.into_iter().last() {
+        if highest_backref > analyzer.group_ix - start_group
+            || highest_backref == analyzer.group_ix && start_group == 0
```
robinst, line 254: Can you add a comment here explaining why `start_group == 0` is special?

src/parse.rs
```diff
@@ -1945,8 +1945,20 @@ mod tests {
     #[test]
     fn invalid_backref() {
         // only syntactic tests; see similar test in analyze module
-        fail(".\\12345678"); // unreasonably large number
```
robinst, line 1948: This was probably meant to be a number that is outside of e.g. usize range, right? We can change it to be a bigger number so that's actually being tested.

src/parse.rs
```diff
@@ -1945,8 +1945,20 @@ mod tests {
...
+        assert_error(".\\c", "Parsing error at position 1: Invalid escape: \\c"); // not decimal
+
+        assert_error(
+            "a\\1",
+            "Parsing error at position 2: Invalid back reference",
+        ); // also an unreasonably large number according to regex length
+        assert_error(
+            "(a)\\2",
+            "Parsing error at position 4: Invalid back reference",
+        ); // also an unreasonably large number according to regex length
```
robinst: I wouldn't call these "unreasonably large" :). Just invalid

robinst (approved):
LGTM, just some comments

src/parse.rs
keith-hall: reworded :)

src/parse.rs
keith-hall: done

src/analyze.rs
keith-hall: good idea, done

PR #179, opened 2025-08-11
Fix behavior of repetition on an empty match

### LINKED ISSUE #70: (|a)* matches more text than (|a)+ does
See https://github.com/rust-lang/regex/issues/779

We might have the same problem in our VM. Need to check what our behavior is (for hard regexes) and adjust it to match the new logic of regex crate, go regex, RE2 and co.

keith-hall:
Previously it would fail the branch when a repetition resulted in an empty match, but Oniguruma treats it as success if enough iterations were completed and breaks out the repetition. So the change is to match Oniguruma behavior. Fixes #70 

I also noticed a test case which after the trailing lookahead optimizations, wasn't testing what it should, so I fixed it.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -645,7 +645,7 @@ pub(crate) fn run(
                     check,
                 } => {
                     let repcount = state.get(repeat);
-                    if repcount > lo && state.get(check) == ix {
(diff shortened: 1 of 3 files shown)


robinst (approved):
Ok, sounds good!

PR #182, opened 2025-08-22
support const-size backrefs in lookbehinds

keith-hall:
I noticed that backrefs were not supported in lookbehinds, always failing with the `LookBehindNotConst` error, even if the group being backreferenced *was* a constant size. This fixes that by updating the analyzer to store the min size and const size properties of the capture group, for later reference by the backreference.

### DIFF
--- a/src/analyze.rs
+++ b/src/analyze.rs
@@ -30,6 +30,11 @@ use crate::alloc::string::ToString;
 use crate::parse::ExprTree;
 use crate::{CompileError, Error, Expr, Result};
 
+#[cfg(not(feature = "std"))]
+use alloc::collections::BTreeMap as Map;
+#[cfg(feature = "std")]
+use std::collections::HashMap as Map;
+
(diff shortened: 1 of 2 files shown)


src/analyze.rs
```diff
@@ -67,6 +72,9 @@ impl<'a> Info<'a> {
 struct Analyzer<'a> {
     backrefs: &'a BitSet,
     group_ix: usize,
+    /// Stores the analysis info for each group by group number
+    // NOTE: uses a Map instead of a Vec because sometimes we start from capture group 1 othertimes 0
+    group_info: Map<usize, (usize, bool)>, // (min_size, const_size)
```
robinst: How about making `(usize, bool)` a struct instead? Then we don't need the `// (min_size, const_size)` comment to say what they mean.

keith-hall: Good idea, done 👍

PR #183, opened 2025-08-22
Fix analyzer start group in optimized patterns in toy example

keith-hall:


keith-hall:
This will already be fixed as part of #181 in a slightly different way

robinst (approved):

Issue #133, opened 2024-01-27
Export `parse` related utilities


LeoDog896:
`fancy-regex` is the only rust crate that has generic regex syntax support; exporting the parse utilities / bundling it as its own crate would be quite nice.

keith-hall:
FYI, this is something we plan to get to sometime after we support subexpressions and recursion. It will probably take some work to get the API in a good enough and stable enough state to make public.

PR #186, opened 2025-09-06
feat: Add an option to better match Oniguruma

### LINKED ISSUE #180: `\>` no longer matches a literal `>`, unlike Oniguruma
It looks like starting with `fancy-regex` version `0.13.0` the regex string `\>` (an incorrectly escaped `>`) no longer matches a literal `>`. This behavior appears to deviate from Oniguruma

## Demo

```toml
[package]
name = "fancy-bug"
version = "0.1.0"
edition = "2024"

[dependencies]
fancy-regex-12 = { package = "fancy-regex", version = "0.12.0" }
fancy-regex-13 = { package = "fancy-regex", version = "0.13.0" }
onig = "6.5.1"
```

```rust
fn main() {
    try_main().unwrap();
}

fn try_main() -> Option<()> {
    let regex_str = r"\>";
    let haystack = " >";

    // fancy-regex 0.12.0
    let fancy_regex_12 = fancy_regex_12::Regex::new(regex_str).ok()?;
    let fancy_pos_12 = fancy_regex_12.captures(haystack).ok()??.get(0)?.range();

    // fancy-regex 0.13.0
    let fancy_regex_13 = fancy_regex_13::Regex::new(regex_str).ok()?;
    // no capture :/
    assert!(fancy_regex_13.captures(haystack).ok()?.is_none());

    // onig latest
    let onig_regex = onig::Regex::new(regex_str).ok()?;
    let onig_pos = onig_regex.captures(haystack)?.pos(0)?;

    println!("fancy-regex 0.12.0: {fancy_pos_12:?}\nonig: {onig_pos:?}");

    Some(())
}
```

```console
$ cargo r
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/fancy-bug`
fancy-regex 0.12.0: 1..2
onig: (1, 2)
```

## Context

This is coming from this `syntect` PR where updating `fancy-regex` led to some issues due to several syntax definitions containing regexes that incorrectly escape `<` and `>`

- trishume/syntect#596

and a relevant comment from @keith-hall as well

- https://github.com/trishume/syntect/pull/596#issuecomment-3182226397

As mentioned in that thread I'm planning on trying to tackle this issue if I can manage it 😼

CosmicHorrorDev:
- Closes #180

This PR adds a method to toggle interpreting `\<` and `\>` as matching the literals `<` and `>` instead of left and right word bounds. Opening as a draft so that naming can be sorted out, but otherwise I'm happy with the changes!

---

`syntect`'s tests all seem to pass with this option enabled :tada: 

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 73351ac..41f8726 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ features = ["metadata"]
 [dependencies]
 yaml-rust = { version = "0.4.5", optional = true }
 onig = { version = "6.5.1", optional = true, default-features = false }
-fancy-regex = { version = "0.11", optional = true }
+fancy-regex = { path = "../../fancy-regex/fancy-regex", optional = true }
 walkdir = "2.0"
 regex-syntax = { version = "0.8", optional = true }
 plist = { version = "1.3", optional = true }
diff --git a/src/parsing/regex.rs b/src/parsing/regex.rs
index bb0500f..990392d 100644
--- a/src/parsing/regex.rs
+++ b/src/parsing/regex.rs
@@ -211,7 +211,7 @@ mod regex_impl {
 
     impl Regex {
         pub fn new(regex_str: &str) -> Result<Regex, Box<dyn Error + Send + Sync + 'static>> {
-            let result = fancy_regex::Regex::new(regex_str);
+            let result = fancy_regex::RegexBuilder::new(regex_str).bikeshed_literals_instead_of_wordbounds(true).build();
             match result {
                 Ok(regex) => Ok(Regex { regex }),
                 Err(error) => Err(Box::new(error)),
```

```console
$ make syntest-fancy
    ... lots of build output
    Finished `release` profile [optimized] target(s) in 3.61s
     Running `target/release/examples/syntest testdata/Packages testdata/Packages --summary`
No new failures!
```


keith-hall:
As Oniguruma doesn't have such an option, we can't get inspired by naming there. So I would ask, do we think that we will want lots of little flags for various things like these, or a more umbrella-like flag encompassing more things, like regex crate syntax compatibility? Because at the end of the day, "Oniguruma-mode" would be to treat these escapes as literals, "regex-crate-mode" would be the half word boundaries.

Can we add some parser tests also please, which would include ``\>`` and ``\<`` in the pattern being parsed, with the new flag enabled and disabled.

src/flags.rs
keith-hall, line 1: I'm wondering if it would make sense to rename this module to `parse_flags` to reflect that the flags only affect the parser, so that once we implement https://github.com/fancy-regex/fancy-regex/issues/133, and it becomes possible to construct a `Regex` from an `ExprTree`, it would be clearer that these flags would have no further effect.
(If we introduce an Oniguruma mode flag, it should also only affect the parser, at least at this stage. So then the regex builder method could become ``parse_oniguruma_syntax`` or ``oniguruma_parse_mode`` or something.
I guess my point is that I don't want it to be confusing for people, thinking that there are multiple VM modes - oniguruma vs regex-crate-with-fancy-features etc.)

CosmicHorrorDev: I'm for that! It took me a good bit to realize that the flags only impact the parser

CosmicHorrorDev:
Renamed things, added another test, and fleshed out more of the docs. Should be all ready for review now :tada:

keith-hall (approved):
Looks good to me, I appreciate the thorough documentation 👍 thanks

keith-hall:
Yep, it's a bug in the code coverage tool, it unfortunately shows false negatives quite a lot with Rust code

src/lib.rs
```diff
@@ -696,6 +700,41 @@ impl RegexBuilder {
...
     }
+
+    /// Attempts to better match [Oniguruma](https://github.com/kkos/oniguruma)'s default behavior
+    ///
+    /// Currently this amounts to changing behavior with:
+    ///
+    /// # Left and right word bounds
+    ///
+    /// `fancy-regex` follows the default of other regex engines such as the `regex` crate itself
+    /// where `\<` and `/>` correspond to a _left_ and _right_ word-bound respectively. This
```
robinst: typo

```suggestion
    /// where `\<` and `\>` correspond to a _left_ and _right_ word-bound respectively. This
```

src/flags.rs
robinst: +1 for the rename

src/lib.rs
CosmicHorrorDev: Good find, fixed!

src/flags.rs
```diff
@@ -4,3 +4,4 @@ pub const FLAG_DOTNL: u32 = 1 << 2;
 pub const FLAG_SWAP_GREED: u32 = 1 << 3;
 pub const FLAG_IGNORE_SPACE: u32 = 1 << 4;
 pub const FLAG_UNICODE: u32 = 1 << 5;
+pub const FLAG_ONIG_MODE: u32 = 1 << 6;
```
keith-hall: I've been thinking about this, and I realized that I'm not a huge fan of abbreviations, because they make it harder to search the codebase for something - i.e. if we search for `Oniguruma`, this won't show up. What do others think about using the full name? `FLAG_ONIGURUMA_MODE`  here and [`oniguruma_mode` in the `RegexOptions` struct](https://github.com/fancy-regex/fancy-regex/pull/186/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R552)?

CosmicHorrorDev: I'm totally fine with that! Feel free to push that change if you want. I won't be able to get to things myself till later next week

src/flags.rs
keith-hall: I pushed a commit performing the rename :)

src/flags.rs
keith-hall: done, thanks :) we have a web-based playground now, I haven't exposed this flag there, but if anybody thinks we should then please let me know or just make a PR :)

PR #192, opened 2025-09-18
Fix clippy lints

keith-hall:
I had the idea that it might be a good idea to fix clippy lints before making the next release.

I also did a [`cargo semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) and it confirmed that the next release can be minor, even after these changes.

Indeed after fixing these, we get some new lint warnings but I think we can leave them for a future release as they will be breaking:
```
warning: the `Err`-variant returned from this function is very large
   --> src/vm.rs:496:57
    |
496 | ...&str, pos: usize) -> Result<Option<Vec<usize>>> {
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
   ::: src/error.rs:17:5
    |
17  |     CompileError(CompileError),
    |     -------------------------- the largest variant contains at least 136 bytes
    |
    = help: try reducing the size of `error::Error`, for example by boxing large elements or replacing it with `Box<error::Error>`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -191,7 +191,7 @@ use core::convert::TryFrom;
 use core::fmt::{Debug, Formatter};
 use core::ops::{Index, Range};
 use core::str::FromStr;
-use core::{fmt, usize};
+use core::{fmt};
 use regex_automata::meta::Regex as RaRegex;
 use regex_automata::util::captures::Captures as RaCaptures;
 use regex_automata::util::syntax::Config as SyntaxConfig;
@@ -1085,7 +1085,7 @@ impl Regex {
     }
 
     /// Returns an iterator over the capture names.
-    pub fn capture_names(&self) -> CaptureNames {
+    pub fn capture_names(&self) -> CaptureNames<'_> {
         let mut names = Vec::new();
         names.resize(self.captures_len(), None);
         for (name, &i) in self.named_groups.iter() {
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -1045,7 +1043,7 @@ impl Regex {
             } => {
                 let mut locations = inner.create_captures();
                 inner.captures(RaInput::new(text).span(pos..text.len()), &mut locations);
-                Ok(locations.is_match().then(|| Captures {
+                Ok(locations.is_match().then_some(Captures {
```
robinst: Hm not sure about this one, won't that eagerly evaluate the argument? I guess there's no calculations in there, but still..

Maybe just rewrite it like this:

```rust
if locations.is_match() {
    Ok(Captures { ... })
} else {
    Ok(None)
}
```

src/parse.rs
```diff
@@ -1146,7 +1143,7 @@ fn is_id_char(c: char) -> bool {
 }
 
 fn is_digit(b: u8) -> bool {
```
robinst: I think we can inline this now?

robinst (approved):
Some suggestions, otherwise LGTM, thanks!

src/lib.rs
keith-hall: I was thinking the same thing, done :)

src/parse.rs
keith-hall: we sure can - done

PR #193, opened 2025-09-28
Enhance word boundary for #136

### LINKED ISSUE #136: Missing regex special word boundary escapes
Specifically, `\b{start}`, `\b{end}`, `\b{start-half}`, `\b{end-half}`. fancy-regex errors when encountering one of these with `InvalidEscape("\\b")`. The directional boundary assertions are available as `\<`/`\>`, but the half-boundary assertions aren't accessible except via lookaround (`(?<=\W|\A)`/`(?=\W|\z)`).

k94-ishi:
closes #136

This PR enhances the handling of `\b` word boundaries.

## Added syntax

- `\b{start}`: Left word boundary (equivalent to `\<`)
- `\b{end}`: Right word boundary (equivalent to `\>`)
- `\b{start-half}`: Left word half boundary
- `\b{end-half}`: Right word half boundary

## Implementation details

- Extended the parser to support brace-style word boundary cases
- Added new `Assertion` variants to represent the half-boundary semantics
  - `Assertion::LeftWordHalfBoundary`
  - `Assertion::RightWordHalfBoundary`

## Notes

- Provides clear error messages for unsupported `\B{...}` syntaxes
- Adds tests for boundary brace parsing and matching
  - Parsing validation tests for valid and invalid syntax
  - Functional tests to verify boundary semantics


tests/finding.rs
```diff
@@ -411,3 +411,26 @@ fn incomplete_escape_sequences() {
...
+    assert!(Regex::new(r"\b{ }").is_err());
+    assert!(Regex::new(r"\b{START}").is_err());
+    assert!(Regex::new(r"\b{END}").is_err());
+
+    // \B{...} -> Fail
+    assert!(Regex::new(r"\B{start}").is_err());
+    assert!(Regex::new(r"\B{end}").is_err());
+    assert!(Regex::new(r"\B{start-half}").is_err());
+    assert!(Regex::new(r"\B{end-half}").is_err());
+}
```
keith-hall: I would prefer to have parser tests in `parse.rs` for consistency - probably the `incomplete_escape_sequences` test should also be moved and reworked slightly. It would also be nice to assert on the `ParseError::InvalidEscape` and check the message, to help prevent regressions in future, compared to just checking for any error.

tests/matching.rs
```diff
@@ -199,6 +199,54 @@ fn hard_trailing_positive_lookaheads() {
...
+    assert_no_match(r"-1-\b{end}", "chapter -1-");
+
+    // \b{end-half}
+    assert_match(r"world\b{end-half}", "hello world");
+    assert_no_match(r"world\b{end-half}", "worldhello");
+    assert_match(r"test\b{end-half}", "run test");
+    assert_no_match(r"test\b{end-half}", "testing");
+    assert_match(r"-1-\b{end-half}", "chapter -1-");
+
+    // \b{start} <=> \< and \b{end} <=> \<
```
keith-hall: all these angle brackets get confusing, can we perhaps use Markdown styling to hint at the equations please? :slightly_smiling_face: 
(it wasn't immediately obvious to me that "<=>" meant "equivalent to")
```suggestion
    // `\b{start}` <=> `\<` and `\b{end}` <=> `\>`
```

src/parse.rs
```diff
@@ -689,6 +681,62 @@ impl<'a> Parser<'a> {
...
+        let bytes = self.re.as_bytes();
+
+        // Find the opening brace (should be at ix + 2, after \b or \B)
+        let start_brace = ix + 2;
+        if bytes.get(start_brace) != Some(&b'{') {
+            return Err(Error::ParseError(
+                ix,
+                ParseError::InvalidEscape("\\b{...}".to_string()),
+            ));
+        }
```
keith-hall: This is currently never going to happen, but I guess it makes sense to leave it as defensive programming for future refactorings...

src/parse.rs
```diff
@@ -689,6 +681,62 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn parse_word_boundary_brace(&self, ix: usize) -> Result<(usize, Expr)> {
```
keith-hall, line 700: maybe a comment hinting where ix is pointing to could be helpful, similar to how we have done on other methods like `parse_hex`, `parse_flags` etc.

keith-hall (commented):
Thanks, looks good. I have a few small ideas for improvement.

src/parse.rs
CAD97: The usual pattern is to `assert!` for things that should always be true by construction, but since emitting a proper error here is nearly as trivial as the panic, doing so seems reasonable.

However, only the `{` and not the `\b` is checked currently, which feels a little odd.

Assert:

```rust
        assert_matches!(
            bytes.get(ix..ix + 3),
            Some([b'\\', b'b' | 'B', b'{']),
            "ix should be at \\b{{...}}",
        );
```

Error:

```rust
        if !matches!(bytes.get(ix..ix + 3), Some([b'\\', b'b' | 'B', b'{'])) {
            return Err(Error::ParseError(
                ix,
                ParseError::InvalidEscape("\\b{...}".to_string()),
            );
        }
```

tests/finding.rs
k94-ishi: Thank you! I’ve updated it as suggested: moved the parsing tests to parse.rs and added assertions on the error message.

src/parse.rs
k94-ishi: @CAD97 Thank you for the suggestion! I’ve updated the code to use error return instead of assert!, following your example.

tests/parse.rs
keith-hall, line 1: Just curious, under what circumstances do we prefer having a separate test file compared to having the test be a part of the unit being tested? i.e. https://github.com/fancy-regex/fancy-regex/blob/main/src/parse.rs#L1161 ? Like having separate tests, we use public API I guess, having them in the module being tested can use private APIs and get closer to the raw details...

tests/parse.rs
```diff
@@ -0,0 +1,61 @@
+use fancy_regex::Regex;
+
+#[test]
+fn word_boundary_brace_parsing_ok() {
+    // Valid patterns should succeed
+    assert!(Regex::new(r"\b{start}").is_ok());
+    assert!(Regex::new(r"\b{end}").is_ok());
+    assert!(Regex::new(r"\b{start-half}").is_ok());
+    assert!(Regex::new(r"\b{end-half}").is_ok());
```
keith-hall: if these were in the parser itself's `test` (sub)module as linked above, we could assert that parsing `\b{start}` produces an `Expr::Assertion(Assertion::LeftWordBoundary)` as opposed to just that constructing a regex with this pattern succeeds. I feel like it would be more beneficial, what do others think?

keith-hall (commented):
Thanks for the updates. I have a couple of more questions/considerations - this may be where my lack of Rust experience shows, so please feel free to tell me that the way it is done now is better than my proposal :)

keith-hall:
One more quick thing, in our [contributing guide](https://github.com/fancy-regex/fancy-regex/blob/main/CONTRIBUTING.md#contributing) we mention keeping the documentation comments in lib.rs up to date with new syntax. It would be great to see that as well please 😎

k94-ishi:
I've moved the tests that were in `tests/parse.rs` into corresponding unit tests in `src/parse.rs`, and then removed `tests/parse.rs`. Since parsing is just one internal step of regex processing, it fits better as a unit test target rather than an integration test.
As for the docs in lib.rs, I checked and since this syntax already exists in the regex crate, I think no update is needed — but happy to add one if you prefer.

keith-hall (approved):
Looks great, thank you!

k94-ishi:
These syntaxes were requested in #136, and they come from the Rust `regex` crate — they’re not specific to Oniguruma.
[https://docs.rs/regex/1.11.3/regex/#empty-matches](https://docs.rs/regex/1.11.3/regex/#empty-matches:~:text=exactly%20n%20x-,%C2%A7,-Empty%20matches)

Since the existing documentation in `lib.rs` already mentions that fancy-regex supports the `regex` crate’s syntax, I think no update is needed.
But I can add a note if needed.

src/parse.rs
```diff
@@ -689,6 +681,62 @@ impl<'a> Parser<'a> {
...
+                ParseError::InvalidEscape(format!("\\B{{{}}}", content)),
+            ));
+        }
+
+        let expr = match content {
+            "start" => Expr::Assertion(Assertion::LeftWordBoundary),
+            "end" => Expr::Assertion(Assertion::RightWordBoundary),
+            "start-half" => Expr::Assertion(Assertion::LeftWordHalfBoundary),
+            "end-half" => Expr::Assertion(Assertion::RightWordHalfBoundary),
+            _ => {
```
robinst, line 765: Note that the regex crate accepts e.g. `\b{1}` or even `\B{1}` or `\b{2}`, so I think this still needs some tweaking.

keith-hall: FWIW this isn't mentioned in the regex crate docs: https://docs.rs/regex/latest/regex/#empty-matches

Trying it out at https://rustexp.lpil.uk/ just seems to indicate that it is repetition, which makes no sense for zero width assertions. I.e. an expression like `\b{9,10}\w` matches `H`. So maybe we can merge this PR as is?

BurntSushi: Yes, it is just repetition. And indeed, adding repetition operators (that require at least one match) to look-around assertions is effectively meaningless. Indeed, `\b\b\b\b\b\b` is just `\b`. And `^$^$^$^$` is just `^$` and is also `$^`.

keith-hall: Thanks for the confirmation 🙂 [fancy-regex will skip repetition after an empty match](https://github.com/fancy-regex/fancy-regex/pull/179), so there is no harm supporting it if we want better compatibility with the regex crate. But we should gate this behavior so that when behind the Oniguruma parse flag, it would behave like Oniguruma and return an error that the target of the repeat operator is invalid

k94-ishi:
I've implemented the changes based on your feedback. Here’s a summary of the updates:

1. Updated `is_repeatable()` to check Oniguruma flag:
   - Default mode: assertions can be repeated (e.g. `\b{1}`, `\B{2}`, `^*`, etc.)
   - Oniguruma mode: returns error "Target of repeat operator is invalid"
2. Enhanced `\b{...}` parsing to distinguish between:
   - Repetition: `\b{1}` - checks for digits or commas after `{`
   - Named boundaries: \b{start} - continues to work as before
3. Updated tests:
   - Added tests for both modes
   - Removed old tests that expected assertion repetition to be an error

Please let me know if anything looks off or could be improved!

Note: The slight coverage decrease is expected, as some old tests that previously expected assertion repetition to be an error were removed.

keith-hall:
> Note: The slight coverage decrease is expected, as some old tests that previously expected assertion repetition to be an error were removed.

Could we not keep them with the Oniguruma mode flag? :slightly_smiling_face: Sometimes unit tests can be really helpful, those were helping to validate that the error location reporting was accurate, that each different repetition type is picked up etc.

I think we need a few small tweaks for "extended" mode, the regex crate seems to strip whitespace inside the curly braces, but fancy-regex in this PR currently reports an error.
```regex
(?x)\b{ s t a r t}\w
```
[image](https://rustexp.lpil.uk/)
fancy-regex parse error

k94-ishi:
You’re absolutely right — I’ve reintroduced those tests under the Oniguruma mode.  

I also fixed the behavior under `FLAG_IGNORE_SPACE` (extended mode):  
`\b{...}`, `\x{...}`, `\u{...}`, and `\U{...}` now correctly skip whitespace and comments inside braces, matching the `regex` crate’s behavior.  

`(?x: \\b { s t a r t })` → parsed as `\\b{start}`  
`(?x: \\b { s t a r t # \n})` → parsed as `\\b{start}`  
`(?x: \\u { 0 0 3 2 })` → parsed as `\\u{0032}`  
`(?x: \\u { 0 0 3 2 # \n })` → parsed as `\\u{0032}`  

I used the function `optional_whitespace` to handle this, so some syntaxes behave slightly differently from the `regex` crate.  

`(?x: \\b { s t a r t (?#)})` → parsed as `\\b{start}`  

Please take a look when you get a chance!

k94-ishi:
In the `regex` crate’s behavior,

- `(?x: \\b { s t a r t })`

```plaintext
Syntax(
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    (?x: \b { s t a r t })
              ^
error: repetition quantifier expects a valid decimal
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
)
```

- `(?x: \\b{ s t a r t (?#)})`

``` plaintext
Syntax(
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    (?x: \b{ s t a r t (?#)})
           ^^^^^^^^^^^^
error: special word boundary assertion is either unclosed or contains an invalid character
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
)
```

keith-hall (approved):
Awesome, thanks very much for your hard work on this one!

keith-hall:
> If these cases should instead raise errors to match the `regex` crate’s behavior, I’ll update them accordingly.

I think trying to exactly match another engine's behavior in all circumstances is an impossible goal. I'm happy with how it is now - the important thing is that all constructs which the regex crate allows are also allowed by fancy-regex when not in Oniguruma mode, for compatibility. If we don't error when regex does (i.e. fancy-regex is more lenient), I think it is fine.

PR #194, opened 2025-10-15
basic variable-sized lookbehind implementation

### LINKED ISSUE #74: Support look-behind assertions without constant size
I've been playing around with Wikipedia's RegExTypoFix project: https://en.wikipedia.org/wiki/Wikipedia:AutoWikiBrowser/Typos

While most of the rules compile with fancy-regex, there are a few hundred that fail with error "Look-behind assertion without constant size", you can try it out at https://github.com/intgr/topy-rs

Not sure how difficult the implementation would be.

keith-hall:
use regex-automata's DFA reverse search functionality to support variable (even unbounded) length lookbehinds without fancy features or capture groups. Partially fixes #74 

Added benchmarks demonstrating that const-size lookbehinds are more efficient:
```
const_size_lookbehind    (~154 ns)  - uses simple GoBack instruction
variable_size_lookbehind (~448 ns)  - uses reverse DFA search
```
The variable-size implementation is about 3x slower but still very fast (sub-microsecond), confirming that the const-size optimization continues to work correctly and is worth keeping.

I created a feature flag for this, but I wasn't sure whether it should be part of the default features or not. [Oniguruma does apparently support this](https://github.com/kkos/oniguruma/blob/master/doc/SYNTAX.md#11-onig_syn_variable_len_look_behind-a), so I included it in the default features for now, but if someone has arguments for removing it, I'm happy to hear them.

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -774,10 +774,7 @@ pub(crate) fn run(
                     let mut cache = cache.borrow_mut();
                     let input = Input::new(s).anchored(Anchored::Yes).range(0..ix);
 
-                    let found_match = match dfa.try_search_rev(&mut cache, &input) {
-                        Ok(Some(_)) => true,
-                        _ => false,
-                    };
+                    let found_match = matches!(dfa.try_search_rev(&mut cache, &input), Ok(Some(_)));
 
                     if !found_match {
                         break 'fail;


src/vm.rs
```diff
@@ -131,6 +131,34 @@ impl core::fmt::Debug for Delegate {
     }
 }
 
+#[cfg(feature = "variable-lookbehinds")]
+#[derive(Clone)]
+/// Delegate matching in reverse to regex-automata
+pub struct ReverseSearch {
```
keith-hall: Thinking about it, I'm not sure I like this naming. It is essentially a Delegate matched backwards/in reverse. Nothing to do with searching 🙂

src/vm.rs
```diff
@@ -733,6 +764,25 @@ pub(crate) fn run(
...
+                    pattern: _,
+                }) => {
+                    // Use regex-automata to search backwards from current position
+                    let mut cache = cache.borrow_mut();
+                    let input = Input::new(s).anchored(Anchored::Yes).range(0..ix);
+
+                    let found_match = match dfa.try_search_rev(&mut cache, &input) {
+                        Ok(Some(_)) => true,
+                        _ => false,
+                    };
```
robinst: Could this use `matches!`?

keith-hall: yup, it can - done :+1:

PR #195, opened 2025-10-31
Fix Send/Sync issues related to DFA cache

keith-hall:
I noticed when running `cargo semver-checks` after merging https://github.com/fancy-regex/fancy-regex/pull/194 that the  `RefCell<Cache>` in `ReverseBackwardsDelegate` prevented `Regex`, `Matches` and other structs from implementing Send/Sync, breaking thread-safety guarantees.

This commit replaces `RefCell<Cache>` with regex-automata's `Pool`

`Pool` doesn't implement `Clone`, so we manually implement `Clone` on `ReverseBackwardsDelegate`. Each clone gets its own fresh/independent cache via `dfa.create_cache()`.

To prevent regressions like this in future, some Send/Sync verification tests have been added.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -21,6 +21,8 @@
 //! Compilation of regexes to VM.
 
 use alloc::string::{String, ToString};
+#[cfg(feature = "variable-lookbehinds")]
+use alloc::sync::Arc;
 use alloc::vec::Vec;
 use regex_automata::meta::Regex as RaRegex;
 use regex_automata::meta::{Builder as RaBuilder, Config as RaConfig};
@@ -464,7 +466,7 @@ impl Compiler {
                         .thompson(thompson::Config::new().reverse(true))
                         .build(&pattern)
                     {
-                        Ok(dfa) => dfa,
+                        Ok(dfa) => Arc::new(dfa),
                         Err(e) => {
                             return Err(Error::CompileError(CompileError::DfaBuildError(
                                 e.to_string(),
@@ -473,7 +475,7 @@ impl Compiler {
                     };
 
                     let create: CachePoolFn = alloc::boxed::Box::new({
-                        let dfa = dfa.clone();
+                        let dfa = Arc::clone(&dfa);
                         move || dfa.create_cache()
                     });
                     let cache_pool = Pool::new(create);
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -71,6 +71,7 @@
 
 use alloc::collections::BTreeSet;
 use alloc::string::String;
+use alloc::sync::Arc;
 use alloc::vec;
 use alloc::vec::Vec;
 use regex_automata::meta::Regex;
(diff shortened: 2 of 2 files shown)


src/vm.rs
```diff
@@ -771,10 +791,14 @@ pub(crate) fn run(
                     pattern: _,
                 }) => {
                     // Use regex-automata to search backwards from current position
-                    let mut cache = cache.borrow_mut();
+                    #[cfg(feature = "std")]
+                    let mut cache_guard = cache.lock().unwrap();
+                    #[cfg(not(feature = "std"))]
+                    let mut cache_guard = cache.borrow_mut();
```
robinst: Comment applies to both std and no-std variants: Is this the actual behavior we want? Doesn't locking it here mean that in the `Sync` case (multiple threads using the same regex), only one of them would be able to run this part at the same time (and the other one would block)? (Similarly, for the no-std case, isn't `borrow_mut` just going to fail?)

I was looking through the regex crate's code to see what it does, and it actually has a `Pool` that has docs that go through the options: https://docs.rs/regex-automata/latest/regex_automata/util/pool/struct.Pool.html

Can you look if we can use `Pool` instead? It looks like it might also save us from needing `#[cfg]`.

tests/send_sync.rs
```diff
@@ -0,0 +1,38 @@
...
+    // Create a regex with variable-length lookbehind to ensure
+    // ReverseBackwardsDelegate is Send/Sync when std is available
+    let re = Regex::new(r"(?<=ab+)x").unwrap();
+    assert_send::<Regex>();
+    assert_sync::<Regex>();
+
+    // Verify it actually works
+    assert!(re.is_match("abx").unwrap());
+    assert!(re.is_match("abbx").unwrap());
+    assert!(!re.is_match("ax").unwrap());
```
robinst: Nice tests, thanks. Here, we could add another one which actually uses a few threads to do the matching.

src/vm.rs
keith-hall: Very good point, thanks! I have had a go at switching to `Pool`. :+1: I will want some help with it, I will write comments

src/compile.rs
```diff
@@ -470,11 +472,15 @@ impl Compiler {
                         }
                     };
 
-                    let cache = core::cell::RefCell::new(dfa.create_cache());
+                    let create: CachePoolFn = alloc::boxed::Box::new({
+                        let dfa = dfa.clone();
+                        move || dfa.create_cache()
+                    });
+                    let cache_pool = Pool::new(create);
```
keith-hall, line 481: I don't like that it clones the DFA just to create a cache for it. Do you think it would make sense to deserialize it or is there a better way? https://docs.rs/regex-automata/0.4.13/regex_automata/dfa/index.html#example-deserialize-a-dfa

tests/send_sync.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[cfg(feature = "std")]
+fn main() {
+    use std::thread;
+
+    // Create a regex with variable-length lookbehind
+    let re = Regex::new(r"(?<=ab+)x").unwrap();
+
+    // Clone it to move into thread
+    let re_clone = re.clone();
```
keith-hall: I could use some advice how to make a good threading test, I don't have much experience with threads in Rust yet. It feels like cloning defeats the point here...

keith-hall: Probably the solution is to use Arc, I will try when I get time, taking inspiration from https://github.com/rust-lang/regex/issues/934

src/vm.rs
keith-hall: I think I got it sorted now :)

tests/send_sync.rs
keith-hall: good idea! done

src/compile.rs
keith-hall: fixed

tests/send_sync.rs
keith-hall: done

PR #196, opened 2025-11-08
Support capture groups in variable lookbehinds

keith-hall:
follow up to https://github.com/fancy-regex/fancy-regex/pull/194, to be merged after https://github.com/fancy-regex/fancy-regex/pull/195

- if the variable lookbehind contains capture groups, then after the reverse DFA matches, in order to extract the capture groups, it will execute the Regex again anchored from that match start position until where the lookbehind began.
- if the variable lookbehind contains no capture groups, no additional delegate compilation is performed.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -852,6 +852,21 @@ mod tests {
         assert_matches!(prog[4], End);
     }
 
+    #[test]
+    #[cfg(feature = "variable-lookbehinds")]
+    fn variable_lookbehind_with_required_feature_backref_captures() {
+        // currently hard variable lookbehinds are unsupported.
+        // the backref to a capture group inside the variable lookbehind makes the capture group hard
+        let tree = Expr::parse_tree(r"(?<=a(b+))\1").unwrap();
+        let info = analyze(&tree, false).unwrap();
+        let result = compile(&info, true);
+        assert!(result.is_err());
+        assert_matches!(
+            result.err().unwrap(),
+            Error::CompileError(CompileError::LookBehindNotConst)
+        );
+    }
+
     fn compile_prog(re: &str) -> Vec<Insn> {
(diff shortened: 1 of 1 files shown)


tests/captures.rs
```diff
@@ -50,6 +50,49 @@ fn captures_after_lookbehind() {
...
+    let captures = captures(r"(?<=(?<first>a+)(?<second>b+))x", "aaabbbx");
+    assert_eq!(captures.len(), 3);
+    assert_match(captures.get(0), "x", 6, 7);
+    assert_match(captures.name("first"), "aaa", 0, 3);
+    assert_match(captures.name("second"), "bbb", 3, 6);
+}
+
+#[test]
+#[cfg(feature = "variable-lookbehinds")]
+fn captures_both_inside_and_outside_variable_lookbehind() {
```
robinst, line 86: Nice test!

src/vm.rs
```diff
@@ -153,6 +153,12 @@ pub struct ReverseBackwardsDelegate {
     pub(crate) dfa: Arc<regex_automata::hybrid::dfa::DFA>,
     /// Cache pool for DFA searches
     pub(crate) cache_pool: Pool<regex_automata::hybrid::dfa::Cache, CachePoolFn>,
+    /// The forward regex for capture group extraction
+    pub(crate) capture_group_extraction_inner: Option<Regex>,
+    /// The first group number that this regex captures (if it contains groups)
+    pub(crate) start_group: usize,
+    /// The last group number
+    pub(crate) end_group: usize,
```
robinst, line 161: I think we have this pattern in a few places (`start_group` and `end_group` fields), should we put them into a nice little struct instead? Something like `GroupRange(usize, usize)` (feel free to choose a better name).

robinst (approved):
LGTM, just a suggestion

Issue #178, opened 2025-08-05
Return a compile error for left recursive subroutine calls


keith-hall:
Let's create a new `CompileError` variant for when a left recursive subroutine call is detected, and update the analyzer to emit it when applicable.
https://github.com/kkos/oniguruma/blob/master/doc/RE#L474-L477

PR #198, opened 2025-11-11
Optimize \G when it is at the start of the pattern

### LINKED ISSUE #197: Optimize \G when it is the first instruction
When `\G` appears at the start of a pattern and fails, the VM continues checking every other position in the haystack despite `\G` never being able to match after the first failure. This causes O(n) behavior where O(1) is possible.

One possible fix for this is fairly simple, determine when `\G` is the first `Expr` in the tree, pass that as a boolean down into the `ContinueFromPreviousMatchEnd` instruction and [instead of `break 'fail`](https://github.com/fancy-regex/fancy-regex/blob/f48e5bdbb26e1548ae3a475fb433140be437f1ed/src/vm.rs#L830-L834) it could do an early `return Ok(None)`
I guess it doesn't have to be the exact first Expr in the tree/pattern, as long as there are only const size min size 0 `Expr`s/`Info`s preceding it.
(N.B. worth remembering that a non-nested `|` after it would render this optimization not possible)

keith-hall:
Fixes #197 

It just so happens that the `min_pos_in_group` logic is something I had prepared a while ago while working on https://github.com/fancy-regex/fancy-regex/issues/178

**before:**
```
+Benchmarking continue_from_end_of_prev_match_short_haystack: Warming up for 3.00
+Benchmarking continue_from_end_of_prev_match_short_haystack: Collecting 100 samp
+continue_from_end_of_prev_match_short_haystack
+                        time:   [95.855 ns 96.628 ns 97.575 ns]
+Found 13 outliers among 100 measurements (13.00%)
+  6 (6.00%) high mild
+  7 (7.00%) high severe
+
+Benchmarking continue_from_end_of_prev_match_long_haystack: Warming up for 3.000
+Benchmarking continue_from_end_of_prev_match_long_haystack: Collecting 100 sampl
+continue_from_end_of_prev_match_long_haystack
+                        time:   [123.82 µs 124.99 µs 126.28 µs]
+Found 10 outliers among 100 measurements (10.00%)
+  9 (9.00%) high mild
+  1 (1.00%) high severe
```

**after:**
```
+Benchmarking continue_from_end_of_prev_match_short_haystack: Warming up for 3.00
+Benchmarking continue_from_end_of_prev_match_short_haystack: Collecting 100 samp
+continue_from_end_of_prev_match_short_haystack
+                        time:   [68.663 ns 69.133 ns 69.658 ns]
+                        change: [-29.485% -28.949% -28.399%] (p = 0.00 < 0.05)
+                        Performance has improved.
+Found 16 outliers among 100 measurements (16.00%)
+  5 (5.00%) high mild
+  11 (11.00%) high severe
+
+Benchmarking continue_from_end_of_prev_match_long_haystack: Warming up for 3.000
+Benchmarking continue_from_end_of_prev_match_long_haystack: Collecting 100 sampl
+continue_from_end_of_prev_match_long_haystack
+                        time:   [68.967 ns 69.351 ns 69.779 ns]
+                        change: [-99.945% -99.945% -99.944%] (p = 0.00 < 0.05)
+                        Performance has improved.
+Found 14 outliers among 100 measurements (14.00%)
+  2 (2.00%) low mild
+  4 (4.00%) high mild
+  8 (8.00%) high severe
```

summary: we get a 99.9% speedup on the long haystack, and 28% on the short one

### DIFF
--- a/src/analyze.rs
+++ b/src/analyze.rs
@@ -42,6 +42,8 @@ pub struct Info<'a> {
     pub(crate) capture_groups: CaptureGroupRange,
     pub(crate) min_size: usize,
     pub(crate) const_size: bool,
+    /// Tracks the minimum number of characters that would be consumed in the innermost capture group
+    /// before this expression is matched.
     pub(crate) min_pos_in_group: usize,
     pub(crate) hard: bool,
     pub(crate) expr: &'a Expr,


src/analyze.rs
```diff
@@ -42,6 +42,7 @@ pub struct Info<'a> {
     pub(crate) capture_groups: CaptureGroupRange,
     pub(crate) min_size: usize,
     pub(crate) const_size: bool,
+    pub(crate) min_pos_in_group: usize,
```
robinst, line 47: Can you add some docs here?

keith-hall: done :slightly_smiling_face:

PR #202, opened 2025-12-06
create CaptureGroupRange struct to replace/encapsulate start_group and end_group fields

keith-hall:
resolves the code review comment at https://github.com/fancy-regex/fancy-regex/pull/196#discussion_r2553704636


robinst (approved):
Nice!

PR #203, opened 2025-12-11
Wrap Prog in Arc and remove Clone from Insn and Prog

keith-hall:
This change significantly improves `Regex` cloning performance. Instead of cloning all instructions when cloning a `Regex`, we now use `Arc<Prog>` to share the program between clones. As a result we can remove the `Clone` trait from `Insn` and `Prog`.


robinst (approved):
Nice

Issue #101, opened 2022-09-04
Bad(?) performance with a specific regex with backreferences


intelfx:
Consider this regular expression:
```
(..)\1\1\1$
```

This regular expression should match 4 identical groups of 2 characters at the end of the input string (e. g. `123456789ABABABAB` should match).

When using fancy-regex to match this regular expression against a 40-character string, a single iteration takes almost 6 µs if using the normal `Regex.is_match()` interface whereas manually compiling and running an `Expr` (as it is done in benchmarks) takes ~300 ns.

For comparison, an equivalent RE using only `regex` syntax (`(00000000|...|FFFFFFFF)$`) takes ~170 ns to run against the same 40-character string.

I understand that `Regex` wraps the input RE in a super-expression implementing the "match anywhere" semantics, which probably explains the 20x performance difference, but I'm wondering if this is expected (and optimizing further is a non-goal or a wishlist-level goal) or if this is unexpected/a bug.

<details><summary>bench.rs</summary>
<p>

```
#[macro_use]
extern crate criterion;
extern crate rand;

use criterion::Criterion;
use std::time::Duration;

use fancy_regex::internal::{analyze, compile, run_default};
use fancy_regex::Expr;

use rand::Rng;

struct Hexadecimal;
impl rand::distributions::Distribution<u8> for Hexadecimal {
    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> u8 {
        const RANGE: u32 = 16;
        const GEN_ASCII_STR_CHARSET: &[u8] = b"0123456789ABCDEF";
        // We can pick from 62 characters. This is so close to a power of 2, 64,
        // that we can do better than `Uniform`. Use a simple bitshift and
        // rejection sampling. We do not use a bitmask, because for small RNGs
        // the most significant bits are usually of higher quality.
        loop {
            let var = rng.next_u32() >> (32 - 4);
            if var < RANGE {
                return GEN_ASCII_STR_CHARSET[var as usize];
            }
        }
    }
}

fn check_normal_expr(c: &mut Criterion) {
[...]

PR #215, opened 2025-12-29
Add RegexSet API

keith-hall:
The `RegexSet` API provides efficient matching of multiple regular expression patterns against the same input text (haystack). This is particularly useful for applications like syntax highlighting, where many patterns need to be matched against each line of text.

The design follows fancy-regex's hybrid approach: patterns are analyzed to determine which can be delegated to high-performance DFA ("easy" patterns) and which require backtracking ("hard" patterns). Easy patterns are combined into a single DFA for parallel evaluation, while hard patterns are evaluated individually. This leaves room for further improvements in future, like creating a prefilter easy pattern for each hard pattern (as described at https://github.com/fancy-regex/fancy-regex/issues/101#issuecomment-3146668829) and compiling the hard patterns as anchored, and/or using threads to perform the hard searches.

1. **Performance**: Minimize overhead for repeated matching against different haystacks (e.g., per-line in syntax highlighting)
2. **Priority-based matching**: When multiple patterns match at the same position, the earliest pattern in the set wins
3. **Lazy evaluation**: Only compute what's necessary for each match
4. **Caching**: Reuse match results to avoid redundant computation

This new API is behind a feature flag (`regex-set`) which is disabled by default, to avoid extra compile times for consumers of this library which won't use it.

N.B. CI requires #212


src/regexset.rs
```diff
@@ -0,0 +1,822 @@
...
+    pattern_indices: Vec<usize>,
+}
+
+#[derive(Debug, Clone)]
+struct Pattern {
+    pattern_id: usize,
+    regex: Regex,
+}
+
+impl RegexSet {
```
Keats, line 172: onig also allows replacing a single pattern: https://github.com/rust-onig/rust-onig/pull/210/changes#diff-d7629a1d5a79a670e6e6acc89ef2e65476f6d7d4bc820d2238263b3b0df125efR166-R202 which is needed in textmate because we often update the first/last pattern of a set.

One thing that is not possible in onig and which would be a great speedup (I think) if passing pre-compiled regex (with their pattern if needed) to skip the compilation. This way you can compile every regex only once for a grammar and reuse it. I think something like 20-30% of giallo flamegraph is compiling regex since onig wants to own them.

BurntSushi:
> 2\. **Priority-based matching**: When multiple patterns match at the same position, the earliest pattern in the set wins

One thing to be aware of here is that this isn't how `RegexSet` in the `regex` crate works. In the `regex` crate, `RegexSet` reports _all_ matches. That is, it does an overlapping search. (`regex-automata` does support leftmost-first matching when using multiple patterns though.)

It might make sense to prominently document this difference, as I could otherwise see folks assuming they have the same semantics.

src/regexset.rs
keith-hall: Thanks for the note, I will have a look how we can accommodate that. Just to make sure I understand correctly, there could even be a change to patterns in a `RegexSet` between match iterations, right? i.e. `iter.next(); regexset.replace_pattern(0, new_precompiled_pattern); iter.next();`

Keats: So in practice the code is: https://github.com/getzola/giallo/blob/c3928d030d413350068f1f9c598c52dc03fe7957/src/tokenizer/mod.rs#L393-L398 with the function tweaking the regex in https://github.com/getzola/giallo/blob/c3928d030d413350068f1f9c598c52dc03fe7957/src/tokenizer/anchors.rs#L39-L70

It can happen on every attempt to match a rule as it depends on the current line context and whether they contain a \A or \G flag that we need to remove because it shouldn't be active a that point in the line.

Keats: It might be solved better by being able to disable flags at the regexset level so we don't have to modify anything though but it caused some issues for vscode-textmate: https://github.com/getzola/zola/issues/2758#issuecomment-2854217148

I'm not too sure why, I haven't looked into it

keith-hall: I have reworked it so we can pass in precompiled `Regex`es, but I ran out of time to fix the tests right now, so marking as a draft. It actually helped me make things a lot cleaner 👍
From here it should be easier to add support for adding/removing/replacing regexes in the set. We will have to rebuild the easy DFA if one of them is easy though...

keith-hall: - [x] Probably I will need to change it to take a reference or an Arc to the Regex, so the RegexSet doesn't take ownership

keith-hall: > It might be solved better by being able to disable flags at the regexset level so we don't have to modify anything though but it caused some issues for vscode-textmate

I think this would require us to run all the matching through the VM and would have performance repercussions also as a result

keith-hall: I'm going to push back on the functionality to add/remove patterns from the set. It would complicate the code quite significantly, when it feels like it won't gain much vs constructing a new RegexSet with precompiled regexes. I know that means we won't have the exact API that Oniguruma has, but I think that is okay.
One alternative I am considering is allowing to disable some regexes in the set, so that they get skipped when searching for the next match, but then at least they are all known in advance when the set (and DFA) is built. Then you could build the set with all variants, disable the ones which don't apply, and enable them when they do. Thoughts?

Keats: That would work as well I think, I don't care about the API matching onig at all.

keith-hall: I have the idea that disabling/removing items from the set is going to be quite unique to TextMate syntax logic, and that now `RegexSet` supports being constructed from pre-built `Regex` instances without taking ownership, that it would be better from `fancy-regex`'s maintenance perspective if `giallo` managed this logic itself, constructing new sets when applicable.

Keats: That's fine, I'm already doing that with onig. Do whatever is better for your crate, not for giallo

PR #230, opened 2026-02-17
Subroutine support

keith-hall:
Add support for subroutine calls (including recursion) using Oniguruma syntax `\g<1>`

I have deliberately kept the VM changes minimal for this initial work, and chose to expand the subroutine calls inline in the compiler, up to the recursion depth limit of 20. I also haven't added any unit tests in the compiler to assert how the program looks with recursive calls, to reduce friction and give us more flexibility to change the implementation later when we work on relative recursion level backreferences. Simple subroutine call compilation is tested, and many matching tests have been added, including passing lots of Oniguruma test cases.

Also, huge effort was put into documentation, to explain how subroutines work.


tests/oniguruma/test_utf8_ignore.c
```diff
@@ -118,39 +91,9 @@
...
-  x3("(?<foo>a|\\(\\g<foo>\\))", "((((((((a))))))))", 0, 17, 1);
-
-  // Compile failed: CompileError(FeatureNotYetSupported("Subroutine Call")
-  x2("\\g<bar>|\\zEND(?<bar>.*abc$)", "abcxxxabc", 0, 9);
-
-  // Compile failed: CompileError(FeatureNotYetSupported("Subroutine Call"))
-  x2("\\g<1>|\\zEND(.a.)", "bac", 0, 3);
-
-  // Compile failed: CompileError(FeatureNotYetSupported("Subroutine Call")
+  // Expected group to exist
```
keith-hall, line 94: this failure occurs due to Oniguruma by default ignoring numbered capture groups (treating them as non-capture groups) when named groups are defined, and doesn't reflect subroutine call support being lacking in any way. I made a test specifically to prove this.

tests/oniguruma/test_utf8_ignore.c
```diff
@@ -82,34 +82,7 @@
...
-  x2("\\g<n>(abc|df(?<n>.YZ){2,8}){0}", "XYZ", 0, 3);
-
-  // Compile failed: CompileError(FeatureNotYetSupported("Subroutine Call"))
-  x2("\\A(?<n>(a\\g<n>)|)\\z", "aaaa", 0, 4);
-
-  // Compile failed: CompileError(FeatureNotYetSupported("Subroutine Call")
-  x2("(?<n>|\\g<m>\\g<n>)\\z|\\zEND (?<m>a|(b)\\g<m>)", "bbbbabba", 0, 8);
-
-  // Compile failed: CompileError(FeatureNotYetSupported("Subroutine Call"))
+  // Match found at start 0 and end 1 (expected 2 and 3)
```
keith-hall, line 85: same cause here for the failure

docs/subroutines/2_flags.md
robinst, line 1: File names with spaces feel a bit weird here, maybe just `flags.md` :)?

Also, I see that these are included via `include_str!`. I'm wondering if it would make sense to give them numbers so that in the file tree/when browing the repo (and reading this diff), they also show up in the same order, i.e. `1_intro.md`, `2_flags.md` etc? WDYT?

docs/subroutines/intro.md
```diff
@@ -0,0 +1,52 @@
+# Subroutines: reusable patterns with stable meaning
+
+Subroutines in fancy-regex are not "inline expansions" and not "dynamic calls".
```
robinst: What are "inline expansions" and "dynamic calls"? It would probably make sense to leave that for later, instead of starting with what they aren't.

Also, as an intro to someone who doesn't know subroutines, it's a lot to read without any context. I would start with an example straight away. And I'm not sure that `(\d)\g<1>\1` is a good one, because it already brings the complexity of updating the capture. I think `(?<word>[a-z]+) ... ` is better, but some people might think "I can just use [a-z]+ again instead of using a subroutine".

How about something like this: `(?<num>\d+\.\d+|\d+) x \g<num>`

It's fairly simple and shows how reuse works.

docs/subroutines/intro.md
```diff
@@ -0,0 +1,52 @@
...
+
+Any capture group can become a subroutine - it just needs to be "called".
+
+Think of a subroutine as:
+
+- defined by a capture group
+- executed exactly the way the capture group was originally defined
+- reusable from multiple places
+
+>> Calling a subroutine does not recompile it in the caller's context.
```
robinst: Double blockquote? Also, not sure what this sentence means, sorry :).

docs/subroutines/intro.md
```diff
@@ -0,0 +1,52 @@
...
+## Example
+
+Let's imagine a pattern which will match a digit and capture it into group 1. Then it will call that capture group as a subroutine. Then it will do a backref to group 1.
+
+This will match only when a digit is followed by another digit and then itself.
+
+```rust
+use fancy_regex::Regex;
+
+let re = Regex::new(r"(\d)\g<1>\1").expect("expected compilation to be successful");
```
robinst: To make this less verbose/distracting, can you turn the `expect(...)` into plain `unwrap()` or maybe even `?` (not sure if/how possible in examples).

docs/subroutines/intro.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Side effects
+
+A subroutine call has one side-effect - it updates the capture group position, which affects backref matching etc.
+
+## Example
+
+Let's imagine a pattern which will match a digit and capture it into group 1. Then it will call that capture group as a subroutine. Then it will do a backref to group 1.
+
+This will match only when a digit is followed by another digit and then itself.
```
robinst: "itself" here is not clear. It refers to "another digit", right?

docs/subroutines/left recursion.md
```diff
@@ -0,0 +1,115 @@
...
+fancy-regex rejects any pattern that could recurse without consuming input.
+
+This restriction:
+
+- Prevents infinite recursion
+- Guarantees termination
+- Makes performance predictable
+- Shifts errors from runtime to compile time (note that this refers to when the regex is compiled to a VM, which is not done by the rust compiler, but when the regex is instanciated/parsed from a string)
+
+The result is a regex engine that is both expressive and safe - without relying on fragile heuristics or input-dependent behavior.
```
robinst: I think this last section here is not necessary, it was pretty clear from the previous sections already. (Is this AI btw?)

docs/syntax.md
```diff
@@ -33,7 +33,11 @@ Backreferences:
 `\k<name>`
 : match the exact string that the capture group named *name* matched \
 `(?P=name)`
-: same as `\k<name>` for compatibility with Python, etc.
+: same as `\k<name>` for compatibility with Python, etc. \
+`\g<name>`
+: call the subroutine defined in capture group named *name* \
+`\g<1>`
+: call the subroutine defined in capture group 1
```
robinst: Can you mention the recursion limit here? Some people might not read on to the subroutines section.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ with the exception that 0.x versions can break between minor versions.
 - Add support for more `\p{...}` and `\P{...}` aliases, for Oniguruma compatibility (#207)
 - Add support for word boundaries and zero-length fancy patterns inside variable lookbehinds (at top level) (#216)
 - Add support for `\R` to mean general newline, matching all common line break characters, treating `\r\n` atomically (#220)
+- Add support for subroutine calls (including recursion) `\g<1>` (#230)
```
robinst: Let's also mention the recursion limit here.

robinst (commented):
I've only had time to read the documentation so far, some feedback on that. But in general great to have good docs with lots of examples :).

docs/subroutines/2_flags.md
keith-hall: good thinking, I have removed the spaces and adding the numbering :+1:

docs/subroutines/intro.md
keith-hall: Nice, updated to your simple example and removed the first line. (I think it is covered by `2_flags.md`)

docs/subroutines/intro.md
keith-hall: removed it as it doesn't contain any new information not covered in `2_flags.md`

docs/subroutines/intro.md
keith-hall: done

docs/subroutines/intro.md
keith-hall: clarified the wording

docs/subroutines/left recursion.md
keith-hall: removed the last section. Yes, I used ChatGPT to get me started, as it was quite daunting initially to document it all

docs/syntax.md
keith-hall: done

CHANGELOG.md
keith-hall: done

src/analyze.rs
```diff
@@ -267,6 +267,24 @@ impl<'a> Analyzer<'a> {
...
+                // Check if this backref is inside a recursed group that it refers to
+                // This happens when:
+                // 1. The backref refers to a group we're currently analyzing
+                // 2. That group has a subroutine call to itself
+                if self.analyzing_groups.contains(group) {
+                    // Check if the group has a recursive subroutine call to itself
+                    if let Some(calls) = self.subroutine_calls.get(&group) {
+                        if calls.iter().any(|call| call.target_group == group) {
+                            return Err(Error::CompileError(Box::new(
+                                CompileError::FeatureNotYetSupported(
```
robinst, line 280: Is this "not yet" supported or an error that will stay an error?

src/compile.rs
```diff
@@ -207,10 +225,63 @@ impl Compiler {
...
+                    // For group 0, visit the entire root info
+                    // For other groups, visit the child of the Group expression
+                    if target_group == 0 {
+                        self.visit(target_info, hard)?;
+                    } else {
+                        // Groups should always have at least one child (the group content)
+                        // If empty, this is an error in the analysis phase
+                        if target_info.children.is_empty() {
+                            return Err(Error::CompileError(Box::new(
+                                CompileError::FeatureNotYetSupported(format!(
```
robinst: Same here, this shouldn't be "not yet", right?

src/compile.rs
```diff
@@ -207,10 +225,63 @@ impl Compiler {
...
+                                    "Subroutine call to empty group {}",
+                                    target_group
+                                )),
+                            )));
+                        }
+                        self.b.add(Insn::SaveCaptureGroupStart(target_group));
+                        self.visit(&target_info.children[0], hard)?;
+                        self.b.add(Insn::Save(target_group * 2 + 1));
```
robinst, line 270: Do we need/want a `SaveCaptureGroupEnd`?

src/compile.rs
```diff
@@ -207,10 +225,63 @@ impl Compiler {
...
+                    }
+
+                    // Pop the recursion stack
+                    self.subroutine_recursion_stack.pop();
+                } else {
+                    // The target group doesn't exist (invalid group reference)
+                    // This should have been caught by analysis, but be defensive
+                    return Err(Error::CompileError(Box::new(
+                        CompileError::FeatureNotYetSupported(format!(
```
robinst: (Same comment here.)

README.md
```diff
@@ -138,8 +139,7 @@ creating the excellent regex crate.
 
 ## Authors
 
-The main author is Raph Levien, with many contributions from Robin Stocker
-and Keith Hall.
+The main authors are Raph Levien, Robin Stocker and Keith Hall.
```
robinst, line 142: 🎉

src/compile.rs
```diff
@@ -207,10 +225,63 @@ impl Compiler {
...
-                    CompileError::FeatureNotYetSupported("Subroutine Call".to_string()),
-                )));
+            Expr::SubroutineCall(target_group) => {
+                // Check if we're already expanding this specific group (direct/indirect recursion)
+                let recursion_count = self
+                    .subroutine_recursion_stack
+                    .iter()
+                    .filter(|&&g| g == target_group)
+                    .count();
+                if recursion_count >= MAX_SUBROUTINE_RECURSION_DEPTH {
```
robinst, line 235: A bit confused by this doing `>= 19` but the release notes saying "including recursion up to 20 levels deep". Shouldn't 20 still be ok? Or am I reading it wrong?

robinst (approved):
LGTM, awesome work! Just some minor comments/questions.

src/compile.rs
keith-hall: The first execution of the capture group technically isn't a recursion, but should count towards the recursion depth, so the count is off by one. As it stands, we match Oniguruma behavior, which also has a limit of 20, so I think the documentation is correct, even though it is admittedly a little confusing.

src/analyze.rs
keith-hall: this one is apparently not needed for [Giallo](https://crates.io/crates/giallo) - none of the TextMate syntaxes seem to use this feature in the regex patterns, but Oniguruma seems to support it, so it is something I would like to eventually tackle, for proper (implicit recursion level aware?) backreference support, so we can support matching palindromes that way etc. Until I work on it, I don't entirely understand how it is supposed to work, but I think it makes sense for it to stay as a not yet supported feature error for now

src/compile.rs
keith-hall: not sure, for consistency it makes sense, but at the same time, right now (until we implement recursion level backrefs and see if we need it to behave differently) it would be essentially identical to the Save instruction, except the slot lookup would then be at runtime instead of compile time...

src/compile.rs
keith-hall: thanks, fixed

src/compile.rs
keith-hall: indeed, thanks, fixed

src/compile.rs
keith-hall: (I think Oniguruma counts like this because the initial visit of a capture group via the normal flow is compiled identically to subroutine call visits.)

robinst: > The first execution of the capture group technically isn't a recursion, but should count towards the recursion depth, so the count is off by one.

Aah I see, that's what I was missing.

PR #233, opened 2026-03-18
Add support for Oniguruma's absent repeater

keith-hall:
Add support for Oniguruma's absent repeater, with documentation.

I also updated the playground's `info_to_tree_node` method to show more details about `Absent` `Expr`s and while I was there, made the `summary` on `AnalysisTreeNode` optional.


robinst (approved):
Nice! Didn't know about this syntax :)

PR #236, opened 2026-04-05
Refactor regex options to make it easier to build multiple patterns using the same options

keith-hall:
Refactor regex options to make it easier to build multiple patterns using the same options
And make it clearer what options are actually used by the VM

This will be especially useful when we add a RegexSet feature.

This allows us to maintain a regex crate adjacent API while also improving ergonomics. I kept the documentation in one place and linked to it from the other to aid maintainance.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,6 +79,12 @@ const MAX_RECURSION: usize = 64;
 #[derive(Debug)]
 pub struct RegexBuilder {
     pattern: String,
+    options: RegexOptionsBuilder,
+}
+
+/// A builder for a `Regex` to allow configuring options.
+#[derive(Debug)]
+pub struct RegexOptionsBuilder {
     options: RegexOptions,
 }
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -599,6 +590,90 @@ impl RegexBuilder {
...
+
+    /// Change the pattern to build. Useful when building multiple regexes from
+    /// many patterns.
+    pub fn pattern(&mut self, pattern: String) -> &mut Self {
+        self.pattern = pattern;
+        self
+    }
+
+    /// See [`RegexOptionsBuilder::case_insensitive`]
+    pub fn case_insensitive(&mut self, yes: bool) -> &mut Self {
```
robinst, line 617: Hm, so which one is the main API that consumers should use? Wondering if these should copy the docs (is there a way to do that without repeating them in the source), or whether the docs on `RegexOptionsBuilder` should have the ``See [`RegexBuilder::case_insensitive`]`` instead?

keith-hall: I would say there isn't really a "main" API here, both are equally valid. Implementers coming from the regex crate and only building one regex, or when building multiple regexes with different options, it is natural to use the familiar one. When working with multiple regexes with the same options, the other makes more sense - it all depends on use case really. I'm happy to swap the docs and the links if you think it is best, I don't know how to copy the docs otherwise without `include_str!` - but having a separate file for each option seems a little overkill...

keith-hall: I have quite a few PRs open or soon to be opened which add regex options, so I'm tempted to say, let's merge it like it is now, and we can always bulk update them later once everything is merged if we decide to switch it. Or, perhaps we can use a macro somehow to only define the documentation once but have it appear in both places.

PR #240, opened 2026-04-13
Implement `find_not_empty` option to reject zero-length matches

keith-hall:
- In the VM, a new `OPTION_FIND_NOT_EMPTY` flag is checked when handling the `End` instruction, ignoring the effect of `\K` when present.
- Public API is exposed via `RegexOptionsBuilder::find_not_empty` and `RegexBuilder::find_not_empty`.
- Return `CompileError::PatternCanNeverMatch` for zero-width-only patterns with `find_not_empty`

When `find_not_empty` is enabled, and some nodes are not const size and have min size 0, we cannot `Wrap` the regex but have to treat those nodes as hard.
This is because otherwise if we have a pattern like `(?:|a|b)`, then it would be delegated to regex-automata and the empty branch would always take precedence - we need to use backtracking to try the other branches.

To help identify where the match attempt started, without `\K` interfering, we introduce a new instruction identical to `Split` which is used to bump the haystack position for unanchored searches, but allows us to track the match attempt start position accurately.

The analyze and compile functions now take a struct argument to make the call sites clearer as opposed to passing in booleans and having to look up what they mean.

I also added the new flag to the playground so we can try it out interactively.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ with the exception that 0.x versions can break between minor versions.
 - Add support for the regex crate's `(?R)` CRLF mode flag (#238)
 - Add support for subroutine calls (including recursion up to 20 levels deep, matching Oniguruma behavior) `\g<1>` (#230)
 - Add support for Oniguruma's absent repeater `(?~abc)` (#233)
+- Add support for ignoring empty matches (#240)
 ### Changed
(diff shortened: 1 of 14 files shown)


michaelblyons:
Not really a code review, but I think the AnalyzeContext change was a positive one. The earlier version of this PR had me wondering how many booleans in a method signature a person would be able to keep straight.

PR #242, opened 2026-04-22
parse: accept `(?:)` + quantifier in Oniguruma mode

stefanobaghino:
Part of a series to close divergences between `fancy-regex` (in `oniguruma_mode`) and Oniguruma surfaced while investigating syntect's `syntest` baselines. Downstream integration + baseline regeneration: trishume/syntect#651 (which flips `known_syntest_failures_fancy.txt` to be byte-for-byte identical to `known_syntest_failures.txt` once this stack lands).

## Summary

Oniguruma accepts `(?:)*` / `(?:)+` / `(?:)?` / `(?:){n,m}` as zero-width no-ops; fancy-regex in `oniguruma_mode` previously rejected them as `TargetNotRepeatable`. Real-world trigger: Sublime-syntax backreference substitution produces `(?:)*` at runtime when a backref captures the empty string (e.g. Markdown fenced-code-block closing).

## Fix

`parse.rs`, `parse_flags`: when `(?:...)` body parses to `Expr::Empty` under `FLAG_ONIGURUMA_MODE`, consume trailing quantifier and modifiers inline, return `Expr::Empty`. Orphan quantifiers (`a+++`) still error — relaxation is scoped to explicit empty groups.

## Stack

On top of #240 (base `find_not_empty`). Follow-up: #243 (another Oniguruma-parity fix).

## Tests

`tests/repro_empty_repeat.rs` — default mode still rejects; Oniguruma mode accepts all four quantifier shapes on empty groups and matches the real syntect Markdown pattern.


keith-hall:
Thanks, I reworked it a bit to match the existing style and fix a bug with extended mode and whitespace.

stefanobaghino:
Keeping it as draft until `find_non_empty` is merged to avoid putting too much stuff into it.

keith-hall (approved):

keith-hall:
Oh I forgot to change the target branch to `main` before merging 🤦‍♂️

stefanobaghino:
> Oh I forgot to change the target branch to `main` before merging 🤦‍♂️

If you want to merge _everything_ to `main`, I can retarget the #243 stack to it. Then you can even close #240. 🙂

stefanobaghino:
Heads up: since this was merged into `find_not_empty` and that base branch was then deleted, the commits here never reached `main`. I've re-landed them together with the follow-up pos-past-end fix in #245, targeting `main` directly.

PR #243, opened 2026-04-22
lib: return Ok(None) when search pos exceeds haystack length

stefanobaghino:
Auto-closed when the `find_not_empty` base branch was deleted on upstream (after #240 was merged and #242 was merged into `find_not_empty` by mistake).

Superseded by #245, which targets `main` directly and carries both #242's commits and this PR's own fix.


stefanobaghino:
Auto-closed when `find_not_empty` was deleted. Re-opened as #245 against `main`, bundling #242's commits with this PR's fix.

PR #246, opened 2026-04-23
implement Seek optimization

### LINKED ISSUE #101: Bad(?) performance with a specific regex with backreferences

keith-hall:
fixes #101 

Implements a Seek pre-filter optimization. It derives a regular approximation of the pattern and uses it to skip directly to the earliest/next plausible match position in the haystack, so the backtracking VM only needs to run from that position onward. For patterns where the hard parts leave useful literals or anchors in the approximation - for example, anchored patterns like `(.{5})\1$` - the `Seek` pre-filter can dramatically reduce the number of positions the VM tries.

While the goal is that the seek pattern would always be a conservative over-approximation, and may report false-positive positions but never skip a true match, it is currently disabled by default, and opt-in only, because it isn't battle tested. (For example, right now, it could accidentally skip matches or perform worse with certain patterns and haystack combinations.) If you are willing to try it, please let us know how you get on!

Currently it decides whether seeking for a pattern is worth it by checking for literals or anchors in the approximation pattern. If no such nodes are found, it will resort to the older checking every position in the haystack approach. The function/callback it uses to check the seek pattern is controllable, so feel free to experiment.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -316,13 +316,71 @@ criterion_group!(
     no_seek_backref_in_long_haystack_no_match,
 );
 
+/// Shared logic for the worst-case seek benchmarks.
+///
+/// Compiles `(\d{3})\1` with the given `seek` setting and builds a haystack of
+/// `"1234567890"` repeated 100 times (optionally followed by `"00000"` to place a
+/// match at the very end).  Because every position in the all-digits haystack is
+/// a candidate for `\d{3}`, the seek pre-filter cannot skip any positions — it
+/// becomes pure overhead compared to running without seek.
+fn bench_digit_backref_worst_case(c: &mut Criterion, name: &str, seek: bool, with_match: bool) {
+    let tree = Expr::parse_tree(r"(\d{3})\1").unwrap();
+    let a = analyze(&tree, AnalyzeContext::default()).unwrap();
+    let p = compile(
+        &a,
+        CompileOptions {
+            contains_subroutines: tree.contains_subroutines,
+            seek_filter: if seek {
+                Some(seek_pattern_is_useful)
+            } else {
+                None
+            },
+            ..CompileOptions::default()
+        },
+    )
+    .unwrap();
+    let mut haystack = "1234567890".repeat(100);
+    if with_match {
+        // Append 5 zeros so the string ends with 6 zeros giving a match ("000" + "000")
+        // near the very end, forcing the seek pre-filter to traverse the whole haystack.
+        haystack.push_str("00000");
+    }
+    c.bench_function(name, |b| b.iter(|| run_default(&p, &haystack, 0).unwrap()));
+}
+
+fn seek_digit_backref_worst_case(c: &mut Criterion) {
+    bench_digit_backref_worst_case(c, "seek_digit_backref_worst_case", true, true);
+}
+
+fn seek_digit_backref_worst_case_no_match(c: &mut Criterion) {
+    bench_digit_backref_worst_case(c, "seek_digit_backref_worst_case_no_match", true, false);
+}
+
+fn no_seek_digit_backref_worst_case(c: &mut Criterion) {
+    bench_digit_backref_worst_case(c, "no_seek_digit_backref_worst_case", false, true);
+}
+
+fn no_seek_digit_backref_worst_case_no_match(c: &mut Criterion) {
+    bench_digit_backref_worst_case(c, "no_seek_digit_backref_worst_case_no_match", false, false);
+}
+
+criterion_group!(
+    name = seek_worst_case_benches;
+    config = Criterion::default();
+    targets = seek_digit_backref_worst_case,
+    seek_digit_backref_worst_case_no_match,
+    no_seek_digit_backref_worst_case,
+    no_seek_digit_backref_worst_case_no_match,
+);
+
 #[cfg(feature = "variable-lookbehinds")]
 criterion_main!(
     benches,
     slow_benches,
     lookbehind_benches,
     continue_from_end_of_prev_match_benches,
-    seek_benches
+    seek_benches,
+    seek_worst_case_benches
 );
 
 #[cfg(not(feature = "variable-lookbehinds"))]
@@ -330,5 +388,6 @@ criterion_main!(
     benches,
     slow_benches,
     continue_from_end_of_prev_match_benches,
-    seek_benches
+    seek_benches,
+    seek_worst_case_benches
 );
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -25,8 +25,8 @@ use criterion::Criterion;
 use std::time::Duration;
 
 use fancy_regex::internal::{analyze, compile, run_default, AnalyzeContext, CompileOptions};
-use fancy_regex::Expr;
 use fancy_regex::seek_pattern_is_useful;
+use fancy_regex::Expr;
 use regex::Regex;
 
 fn parse_lifetime_re(c: &mut Criterion) {
@@ -270,7 +270,11 @@ fn bench_backref_in_long_haystack(c: &mut Criterion, name: &str, seek: bool, wit
         &a,
         CompileOptions {
             contains_subroutines: tree.contains_subroutines,
-            seek_filter: if seek { Some(seek_pattern_is_useful) } else { None },
+            seek_filter: if seek {
+                Some(seek_pattern_is_useful)
+            } else {
+                None
+            },
             ..CompileOptions::default()
         },
     )
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -26,6 +26,7 @@ use std::time::Duration;
 
 use fancy_regex::internal::{analyze, compile, run_default, AnalyzeContext, CompileOptions};
 use fancy_regex::Expr;
+use fancy_regex::seek_pattern_is_useful;
 use regex::Regex;
 
 fn parse_lifetime_re(c: &mut Criterion) {
@@ -255,17 +256,75 @@ criterion_group!(
     continue_from_end_of_prev_match_long_haystack,
 );
 
+/// Shared logic for the backref-in-long-haystack seek benchmarks.
+///
+/// Compiles `(abc)\1` with the given `seek` setting, builds a 10,000-`x` haystack
+/// (optionally with `"abcabc"` appended), and registers a criterion benchmark under
+/// `name`.
+fn bench_backref_in_long_haystack(c: &mut Criterion, name: &str, seek: bool, with_match: bool) {
+    // Pattern with a backref — when seek is true the approximation inlines the group
+    // body ("abc"), so the engine jumps directly to "abc" occurrences and skips the rest.
+    let tree = Expr::parse_tree(r"(abc)\1").unwrap();
+    let a = analyze(&tree, AnalyzeContext::default()).unwrap();
+    let p = compile(
+        &a,
+        CompileOptions {
+            contains_subroutines: tree.contains_subroutines,
+            seek_filter: if seek { Some(seek_pattern_is_useful) } else { None },
+            ..CompileOptions::default()
+        },
+    )
+    .unwrap();
+    let mut haystack = String::new();
+    for _ in 0..10_000 {
+        haystack.push('x');
+    }
+    if with_match {
+        // Place the match target at the very end so the seek pre-filter must scan
+        // the entire haystack before finding the single candidate position.
+        haystack.push_str("abcabc");
+    }
+    c.bench_function(name, |b| b.iter(|| run_default(&p, &haystack, 0).unwrap()));
+}
+
+fn seek_backref_in_long_haystack(c: &mut Criterion) {
+    bench_backref_in_long_haystack(c, "seek_backref_in_long_haystack", true, true);
+}
+
+fn seek_backref_in_long_haystack_no_match(c: &mut Criterion) {
+    bench_backref_in_long_haystack(c, "seek_backref_in_long_haystack_no_match", true, false);
+}
+
+fn no_seek_backref_in_long_haystack(c: &mut Criterion) {
+    bench_backref_in_long_haystack(c, "no_seek_backref_in_long_haystack", false, true);
+}
+
+fn no_seek_backref_in_long_haystack_no_match(c: &mut Criterion) {
+    bench_backref_in_long_haystack(c, "no_seek_backref_in_long_haystack_no_match", false, false);
+}
+
+criterion_group!(
+    name = seek_benches;
+    config = Criterion::default();
+    targets = seek_backref_in_long_haystack,
+    seek_backref_in_long_haystack_no_match,
+    no_seek_backref_in_long_haystack,
+    no_seek_backref_in_long_haystack_no_match,
+);
+
 #[cfg(feature = "variable-lookbehinds")]
 criterion_main!(
     benches,
     slow_benches,
     lookbehind_benches,
-    continue_from_end_of_prev_match_benches
+    continue_from_end_of_prev_match_benches,
+    seek_benches
 );
 
 #[cfg(not(feature = "variable-lookbehinds"))]
 criterion_main!(
     benches,
     slow_benches,
-    continue_from_end_of_prev_match_benches
+    continue_from_end_of_prev_match_benches,
+    seek_benches
 );
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ with the exception that 0.x versions can break between minor versions.
 - Add support for Oniguruma's absent repeater `(?~abc)` (#233)
 - Add support for ignoring empty matches (#240)
 - Add support for treating unnamed capture groups as non-capturing when named groups exist (#241)
+- Add experimental seek optimization to only invoke the VM at candidate positions where a match could occur (#246)
 ### Changed
 - `RegexBuilder` can now build multiple patterns with the same options (#213)
 - Parsing of capture group names is now more lenient (#241)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html),
 with the exception that 0.x versions can break between minor versions.
 
+## [Unreleased]
+### Added
+- Add experimental seek optimization to only invoke the VM at candidate positions where a match could occur (#246)
+### Changed
+### Fixed
+
 ## [0.18.0] - 2026-04-24
 ### Added
 - Add support for `\N` to mean any character except newline, for Oniguruma compatibility (#206)
@@ -18,7 +24,6 @@ with the exception that 0.x versions can break between minor versions.
 - Add support for Oniguruma's absent repeater `(?~abc)` (#233)
 - Add support for ignoring empty matches (#240)
 - Add support for treating unnamed capture groups as non-capturing when named groups exist (#241)
-- Add experimental seek optimization to only invoke the VM at candidate positions where a match could occur (#246)
 ### Changed
 - `RegexBuilder` can now build multiple patterns with the same options (#213)
 - Parsing of capture group names is now more lenient (#241)
--- a/PERFORMANCE.md
+++ b/PERFORMANCE.md
@@ -76,12 +76,23 @@ Oniguruma doesn't have a problem with this particular case because its
 optimization saves it again: It checks if there's a `c` in the input
 before doing any matching.
 
-There's nothing preventing fancy-regex from adding similar optimizations
-in the future, but it's not done yet.
+fancy-regex now has a **Seek** pre-filter optimization (disabled by
+default) for hard (backtracking) patterns, which provides similar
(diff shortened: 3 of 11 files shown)


robinst:
@keith-hall Could you maybe cut a release before merging this? There's a lot of good unreleased changes now that would be nice to put into a release.

keith-hall:
> @keith-hall Could you maybe cut a release before merging this? There's a lot of good unreleased changes now that would be nice to put into a release.

Done
https://github.com/fancy-regex/fancy-regex/releases/tag/0.18.0
https://crates.io/crates/fancy-regex/0.18.0

benches/bench.rs
```diff
@@ -255,17 +256,79 @@ criterion_group!(
...
+}
+
+criterion_group!(
+    name = seek_benches;
+    config = Criterion::default();
+    targets = seek_backref_in_long_haystack,
+    seek_backref_in_long_haystack_no_match,
+    no_seek_backref_in_long_haystack,
+    no_seek_backref_in_long_haystack_no_match,
+);
```
keith-hall, line 317: Some concrete benchmark numbers for the best case scenario:

| Benchmark | seek | Time |
|---|---|---|
| `seek_backref_in_long_haystack` | true | ~355 ns |
| `no_seek_backref_in_long_haystack` | false | ~224 µs |
| `seek_backref_in_long_haystack_no_match` | true | ~318 ns |
| `no_seek_backref_in_long_haystack_no_match` | false | ~216 µs |

The seek pre-filter is **~631× faster** in the match case and **~680× faster** in the no-match case.

keith-hall: Some concrete benchmark numbers for the worst case scenario:

| Benchmark | seek | Time |
|---|---|---|
| `seek_digit_backref_worst_case` | true | ~112 µs |
| `no_seek_digit_backref_worst_case` | false | ~90 µs |
| `seek_digit_backref_worst_case_no_match` | true | ~112 µs |
| `no_seek_digit_backref_worst_case_no_match` | false | ~91 µs |

Here, the seek pre-filter is **~1.2x slower** in both match and no-match cases. This occurs because with digit patterns like `\d{3}`, there are many potential match positions throughout the haystack (every digit could start a 3-digit sequence), so the seek optimization doesn't reduce the search space at all. The overhead of checking for the seek condition outweighs the benefits, making the no-seek approach slightly faster in this scenario.

This is why I feel it is beneficial to have, as part of the public API, a way to control whether seeking is considered useful for a particular pattern. It may make sense to extend the callback to pass in the original pattern also. And perhaps to consider making it dynamic such that the seek filter will fallback to the "split unanchored" approach automatically as opposed to compiling a separate programs depending whether seek is considered useful or not.

keith-hall:
given that the worst case slowdown is 1.2x and the speedups can be massive, I'll merge it. We can always reiterate later, even with breaking changes to the API if we need to - it is marked as experimental after all.

stefanobaghino:
Just wanted to share a real-world data point on this optimization, since the speedup turned out to be much bigger than I expected.

Context: I was diagnosing why [`syntect`](https://github.com/trishume/syntect) (Rust syntax highlighter) is ~4× slower with the `regex-fancy` backend than with `regex-onig` on heavy TypeScript files. I'd assumed the gap was structural — `syntect`'s Sublime grammars use enough lookarounds, atomic groups, `\K`, `\G`, and `\R` that roughly half the compiled patterns land on the Fancy/VM arm by necessity, and PMC counter analysis on Apple Silicon (`xctrace --template "CPU Counters"`) showed the VM was already running at IPC ≈ 5 — i.e. close to peak pipeline saturation, with no big frontend or backend stall to remove. The conclusion looked like "the only way to make it faster is to execute fewer instructions, which means changing what `vm::run` does on the hot path."

Then I synced fancy-regex from upstream and noticed PR #246. Wired it into `syntect` by adding `.seek(true)` to the two `RegexBuilder` calls in `src/parsing/regex.rs` — no other changes — and the gap collapsed.

Measurements with hyperfine, 10 runs each, AC power, M-class P-core, rendering [`microsoft/TypeScript/src/compiler/checker.ts`](https://github.com/microsoft/TypeScript/blob/f350b52331494b68c90ab02e2b6d0828d2a22a74/src/compiler/checker.ts) (3.2 MB) via `syncat`, against `syntect` PR [trishume/syntect#627](https://github.com/trishume/syntect/pull/627):

| build                       | mean      | σ        | vs onig |
| --------------------------- | --------- | -------- | ------- |
| `regex-onig`                | 3.225 s   | 0.048 s  | 1.00×   |
| `regex-fancy` (no seek)     | 14.476 s  | 0.428 s  | 4.49×   |
| **`regex-fancy` `seek(true)`** | **6.288 s** | **0.073 s** | **1.95×** |

That's a **2.30× speedup** over the same pre-filter-disabled fancy build, on a workload where seek wins exactly because it's supposed to: a long haystack, many compiled patterns, most patterns failing to match at most positions, but a non-trivial backtracking VM behind each pattern. Variance also tightened (σ went from 3.0% to 1.2%), which is consistent with the VM running on far fewer haystack positions and so producing more uniform per-input behaviour. Smaller inputs (~250 KB Rust/JS files) showed similar relative wins.

A few things stood out reading the seek module:

* The DFA pre-filter being a conservative over-approximation that never gives false negatives makes this drop-in safe. I swapped `oniguruma_mode(true).seek(true)` into `syntect`, ran the full syntax-test suite (~250 lib tests + every `testdata/Packages/*` snapshot) and it passed unchanged.
* The fact that `seek` *layers on top of* the existing Fancy arm (rather than re-routing patterns to the Wrap arm) is the key insight. The hard patterns in `syntect`'s grammars genuinely need backtracking; they can't be expressed in `regex-automata` directly. But they *can* have a permissive DFA approximation that finds candidate positions cheaply, and that's where almost all of the wall-clock cost was going.
* Not turning seek on by default seems right at this stage, but it'd be worth surfacing in the README/changelog as a "consider this if your hard patterns search large haystacks" knob — anyone in a `syntect`-like situation will likely see similar numbers.

Thanks for the work — this materially closes a gap I'd written off as architectural.

keith-hall:
@stefanobaghino thanks for the feedback and concrete numbers! Hearing that there were no regressions makes me more confident in the DFA prefilter construction - I was worried that there are likely to be many scenarios where it missed real match positions, simply because so many weird and whacky patterns are used in the wild that I was bound to have missed some complex edge case. But perhaps the over conservative approach used at the moment is sound. We'll see what other feedback we get, it would be nice to be able to enable it by default. 

That said, I'm still disappointed that fancy-regex is 1.95x slower than Oniguruma. If we want fancy-regex to be taken seriously, it would really help to close that performance gap. One quick question: was that benchmarking done with fancy-regex compiled in release mode or debug mode? (`cargo bench` enables optimizations by default, but it wasn't clear to me if that is what you used or not.)

stefanobaghino:
@keith-hall It was on release IIRC, but I'll double check. I'm working on a couple of experiments that should improve performance a bit more.

stefanobaghino:
Correction to my earlier message: the suite I ran was `cargo test` (syntect's lib tests + the `html_inputs` HTML-output snapshot tests), not `make syntest-fancy`, which is the harness that exercises per-line scope assertions in `testdata/Packages/*/*.sublime-syntax-test`. With `seek(true)` enabled, that harness produces 3 failures on otherwise-passing baselines — the C++ and Objective-C++ template `<>` lookahead, and a Makefile `(?={{rule_lookahead}})` lookahead. So my "passed unchanged" framing was narrower than it sounded, and the paraphrase as "no regressions" was strictly stronger than what I'd actually verified — apologies for the noise.

Root cause turned out not to be the seek over-approximation itself, but the post-`optimize_trailing_lookahead` interaction with `regex-automata`'s META engine: a top-level `(?=X)` is rewritten to `Concat([Group(Empty), X])`, and feeding the body's approximation to META has been observed to mis-report leftmost position on shapes like `[char-class]+<…(?:nested)?…>` (PikeVM and the lazy DFA both agree on a strictly earlier start). With #249 applied, the syntest harness goes back to 0 failures.

Issue #84, opened 2021-11-30
Support for the `bytes` API?


Alphare:
Hi,

I'm looking into `fancy-regex` to maybe add it to Mercurial's `hgignore` system instead of `regex` to allow for non-linear regex as a fallback. Right now we completely disable the Rust fast-path if we come across an unsupported pattern, which is kind of a bummer.

We're using the `bytes` API from the `regex` crate and it looks like it's missing from `fancy-regex`. Is is something that can be added?

Thanks

robinst:
Hey!

I haven't looked into how the `bytes` API works in `regex` at all, but given that we either delegate to regex or use our own VM which already internally prefers working with bytes, it should be possible. The biggest hurdle would probably be creating and maintaining the parallel API. I think `regex` does it via macros but yeah.

So it's probably not a trivial change, and I don't have any plans to work at it at the moment. I'll label this "help wanted" and see if someone wants to pick it up.

As a compromise for your solution, would it be possible to delegate to fancy-regex just in case your input and patterns are well-formed UTF-8 (which is probably the vast majority of cases)?

Alphare:
Thanks for getting back to me!

I understand this is a decent amount of work, I think falling back to `fancy-regex` if `regex` itself fails (so reversing the fallback) and only doing the UTF8 conversion in that case should be a decent workaround that wouldn't impact performance for the "common case".

I'll keep you updated on that experiment once I get some time to try it out.

Alphare:
I just tried it (something took some time to install :sweat_smile:), and there are a couple of issues with the workaround, I'll document those so it's easier to follow:

- Not being able to pass in bytes means a utf-8 conversion/check for every single call (can be hundreds of thousands or millions per invocation in large repositories)
- Either changing the signature from `path -> bool` to `path -> Result<bool, UTF8Error>` or silently eating those errors when they happen
- `fancy-regex` does not have a way of specifying (for example in `RegexBuilder`) to *not* try `regex` first, which slows us down because we've already tried it and it's doing the work twice
- The performance seems to fall off of a cliff, but that's probably to do with the pattern themselves, which can most likely be turned into linear regex with some effort involved. Another idea I had was to identify patterns that aren't linear and split the linear parts from the backtracking ones to feed the latter to `fancy-regex` to have it build a smaller VM. But that's probably a lot of work.

Thanks for listening ;)

lu-zero:
[...]

PR #248, opened 2026-05-01
feat: add bytes-level matching API

lu-zero:
Add Regex::is_match_bytes() and Regex::find_from_pos_bytes() methods that operate on &[u8] without requiring valid UTF-8. Internally this adds run_bytes and helper functions in the VM for byte-level matching.

The Wrap path delegates to regex_automata which already supports bytes. The Fancy path uses a byte-adapted copy of the backtracking VM.

Attempt at addressing #84

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -159,7 +159,7 @@ impl RegexInput for [u8] {
         MatchBytes::new(self, start, end)
     }
     fn advance_position(&self, i: usize) -> usize {
-        next_input_pos(self, i)
+        i + 1
     }
 }
 
@@ -189,6 +189,6 @@ impl<const N: usize> RegexInput for [u8; N] {
         MatchBytes::new(self, start, end)
     }
     fn advance_position(&self, i: usize) -> usize {
-        next_input_pos(self, i)
+        i + 1
     }
 }
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,0 +1,194 @@
+use crate::bytes::MatchBytes;
+use crate::Match;
+use alloc::string::String;
+
+/// Returns the smallest possible index of the next valid UTF-8 sequence
+/// starting after `i`.
+///
+/// Adapted from a function with the same name in the `regex` crate.
+pub(crate) fn next_input_pos(text: &[u8], i: usize) -> usize {
+    let b = match text.get(i) {
+        None => return i + 1,
+        Some(&b) => b,
+    };
+    i + crate::codepoint_len(b)
+}
+
+/// A trait abstracting over input types for regex matching.
+///
+/// This trait is implemented for `str`, `String`, `[u8]`, and `[u8; N]`, allowing
+/// regex methods to work with both UTF-8 text and raw byte slices.
+///
+/// When the regex is compiled with [`BytesMode::Ascii`](crate::BytesMode),
+/// patterns like `.` will match any byte in `[u8]` input. Without bytes mode,
+/// `.` only matches valid UTF-8 codepoints even in byte slices.
+///
+/// The associated type [`Match<'t>`](Self::Match) is the concrete match type
+/// returned for a given input: [`Match<'t>`](crate::Match) for string inputs,
+/// [`MatchBytes<'t>`](crate::MatchBytes) for byte slice inputs.
+pub trait RegexInput {
+    /// The match type produced for this input.
+    type Match<'t>
+    where
+        Self: 't;
+
+    /// Returns the length of the input in bytes.
+    fn len(&self) -> usize;
+    /// Returns the input as a raw byte slice.
+    fn as_bytes(&self) -> &[u8];
+    /// Returns `true` if `ix` is a valid position between UTF-8 codepoints.
+    fn is_char_boundary(&self, ix: usize) -> bool;
+    /// Returns `true` if the entire input is ASCII.
+    fn is_ascii(&self) -> bool;
+    /// Steps back one codepoint from position `i`, returning the start position.
+    fn prev_codepoint_ix(&self, i: usize) -> usize;
+    /// Returns `true` if the input is empty.
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+    /// Construct a match from byte offsets.
+    fn make_match<'t>(&'t self, start: usize, end: usize) -> Self::Match<'t>;
+    /// Advance past the codepoint at position `i`.
+    fn advance_position(&self, i: usize) -> usize;
+}
+
+impl<S: RegexInput + ?Sized> RegexInput for &S {
+    type Match<'t>
+        = S::Match<'t>
+    where
+        Self: 't;
+
+    fn len(&self) -> usize {
+        (**self).len()
+    }
+    fn as_bytes(&self) -> &[u8] {
+        (**self).as_bytes()
+    }
+    fn is_char_boundary(&self, ix: usize) -> bool {
+        (**self).is_char_boundary(ix)
+    }
+    fn is_ascii(&self) -> bool {
+        (**self).is_ascii()
+    }
+    fn prev_codepoint_ix(&self, i: usize) -> usize {
+        (**self).prev_codepoint_ix(i)
+    }
+    fn make_match<'t>(&'t self, start: usize, end: usize) -> Self::Match<'t> {
+        (**self).make_match(start, end)
+    }
+    fn advance_position(&self, i: usize) -> usize {
+        (**self).advance_position(i)
+    }
+}
+
+impl RegexInput for str {
+    type Match<'t> = Match<'t>;
+
+    fn len(&self) -> usize {
+        str::len(self)
+    }
+    fn as_bytes(&self) -> &[u8] {
+        str::as_bytes(self)
+    }
+    fn is_char_boundary(&self, ix: usize) -> bool {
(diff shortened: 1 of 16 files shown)


src/vm.rs
```diff
@@ -1070,6 +1121,365 @@ pub(crate) fn run(
     }
 }
 
+/// Run the program on a byte slice with options.
+#[allow(clippy::cognitive_complexity)]
+pub(crate) fn run_bytes(
```
lu-zero: This is a quite a bit of code duplication, tomorrow I'll try to make it more compact if the proposed small API extension is acceptable.

src/lib.rs
```diff
@@ -2276,6 +2401,17 @@ fn prev_codepoint_ix(s: &str, mut ix: usize) -> usize {
     ix
 }
 
+// precondition: ix > 0
+fn prev_codepoint_ix_bytes(s: &[u8], mut ix: usize) -> usize {
```
keith-hall: as this method is only used in one place, perhaps it makes sense to move it to `impl RegexInput for [u8]`

src/vm.rs
```diff
@@ -590,7 +626,35 @@ fn matches_literal_casei(s: &str, ix: usize, end: usize, literal: &str) -> bool
...
+        return false;
+    }
+    if matches_literal(s, ix, end, literal) {
+        return true;
+    }
+    if !s.is_char_boundary(ix) || !s.is_char_boundary(end) {
+        return false;
+    }
+    let text_bytes = &s.as_bytes()[ix..end];
+    if text_bytes.iter().all(|&b| b < 128) && literal.iter().all(|&b| b < 128) {
```
keith-hall: I guess there is no way to keep the `is_ascii` call now that we operate on bytes? I liked having that abstracted away...

src/lib.rs
```diff
@@ -987,6 +1022,65 @@ impl Regex {
...
+            RegexImpl::Fancy { prog, options, .. } => {
+                let option_flags = if options.find_not_empty {
+                    OPTION_FIND_NOT_EMPTY
+                } else {
+                    0
+                };
+                let result = vm::run(prog, bytes, pos, option_flags, options)?;
+                Ok(result.map(|saves| MatchBytes::new(bytes, saves[0], saves[1])))
+            }
+        }
```
keith-hall, line 1149: As the logic in this function is pretty much identical to the non-bytes implementation, is there a way we can genericize this also to ease maintenance?

src/vm.rs
```diff
@@ -625,7 +689,12 @@ fn store_capture_groups(
 }
 
 /// Run the program with trace printing for debugging.
-pub fn run_trace(prog: &Prog, s: &str, pos: usize) -> Result<Option<Vec<usize>>> {
+#[allow(private_bounds)]
```
keith-hall: I'm not a super Rust expert, is this generally recommended?

src/vm.rs
```diff
@@ -537,33 +590,16 @@ impl State {
...
-        return true;
-    }
-    if !s.is_char_boundary(ix) || !s.is_char_boundary(end) {
-        return false;
-    }
-    if s[ix..end].is_ascii() {
-        return s[ix..end].eq_ignore_ascii_case(literal);
-    }
-
-    // text captured and being backreferenced is not ascii, so we utilize regex-automata's case insensitive matching
```
keith-hall, line 566: I may be biased because I'm the one whom wrote it, but I think it would be clearer if we could keep this comment, to indicate why we do what we do

keith-hall (commented):
Nice, thanks. A few small questions/improvement suggestions.

src/lib.rs
lu-zero: I can try, I started with the quickest and most minimal set of changes, but if there is interest I can iterate as much as needed.

src/vm.rs
lu-zero: It is an ugly hack. let me drop it and keep the public api compatible. If we agree I can break it I can float the generics up and possibly make everything quite ergonomic.

keith-hall: Sure, go ahead 🙂 thanks

src/lib.rs
```diff
@@ -876,6 +886,31 @@ impl Regex {
...
+    /// # use fancy_regex::Regex;
+    ///
+    /// let re = Regex::new(r"\d+").unwrap();
+    /// assert!(re.is_match_bytes(b"abc 123").unwrap());
+    /// ```
+    pub fn is_match_bytes(&self, bytes: &[u8]) -> Result<bool> {
+        match &self.inner {
+            RegexImpl::Wrap { inner, .. } => Ok(inner.is_match(RaInput::new(bytes))),
+            RegexImpl::Fancy { prog, options, .. } => {
+                let result = vm::run(prog, bytes, 0, 0, options)?;
```
keith-hall: I think here it would be preferable to call `find_from_pos_bytes` instead of directly invoking the VM, to mimic how the `str` `is_match` method works, to ensure flags like "find not empty" are passed in correctly.

keith-hall:
Once the code changes are finalized, it would be great to introduce some tests for the byte apis. I wonder whether it would make sense to somehow adapt all existing test cases to run with both apis for full coverage, or that is overkill and just a smaller set of tests should suffice...

lu-zero:
Now most API surface is now generic, I left out only replace and split.

src/lib.rs
```diff
@@ -170,50 +227,40 @@ impl<'r, 't> Matches<'r, 't> {
...
-                // Stop on first error: If an error is encountered, return it, and set the "last match position"
-                // to the string length, so that the next next() call will return None, to prevent an infinite loop.
                 self.last_end = self.text.len() + 1;
                 return Some(Err(error));
             }
             Ok(None) => return None,
             Ok(Some(pair)) => pair,
         };
 
-        if mat.start == mat.end {
```
keith-hall, line 184: It feels like lots of genuinely useful comments got removed here...

lu-zero: Let me do a pass and restore them all :/ sorry!

lu-zero:
Now all the comments are preserved, the two commits could be folded in one, it is just to split the most needed change to an attempt to make the API as complete as possible.

src/vm.rs
```diff
@@ -590,7 +576,36 @@ fn matches_literal_casei(s: &str, ix: usize, end: usize, literal: &str) -> bool
...
+        return false;
+    }
+    if matches_literal(s, ix, end, literal) {
+        return true;
+    }
+    if !s.is_char_boundary(ix) || !s.is_char_boundary(end) {
+        return false;
+    }
+    let text_bytes = &s.as_bytes()[ix..end];
+    if s.is_ascii() && literal.is_ascii() {
```
keith-hall:
```suggestion
    if s[ix..end].is_ascii() && literal.is_ascii() {
```

src/lib.rs
```diff
@@ -733,6 +861,13 @@ impl RegexBuilder {
         self
     }
 
+    /// See [`RegexOptionsBuilder::bytes_mode`]
+    /// See [`RegexOptionsBuilder::bytes_mode`]
```
keith-hall, line 864:
```suggestion
```

src/lib.rs
```diff
@@ -1476,49 +1651,43 @@ impl<'t> From<Match<'t>> for Range<usize> {
...
+    /// How many groups were captured. This is always at least 1 because group 0 returns the whole
+    /// match.
+    pub fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    /// Returns the byte slice of the input that was matched.
+    pub fn as_bytes(&self) -> &'t [u8] {
+        self.input.as_bytes()
+    }
```
keith-hall, line 1685: `Captures::as_bytes()` is named confusingly vs `MatchBytes::as_bytes()`

`MatchBytes::as_bytes()` returns the matched bytes (`&self.bytes[self.match_start..self.match_end]`). But `Captures::as_bytes()` returns the entire input (`self.input.as_bytes()`). This is a confusing name collision - a user reading the API could easily expect `caps.as_bytes()` to return the matched bytes (equivalent to `caps.get(0).unwrap().as_bytes()`).

Consider renaming to `Captures::input_as_bytes()`  or something equivalently similar to how `CaptureMatches::text()` is named.

src/input.rs
```diff
@@ -0,0 +1,194 @@
...
+        while i > 0 && self[i] & 0xC0 == 0x80 {
+            i -= 1;
+        }
+        i
+    }
+    fn make_match<'t>(&'t self, start: usize, end: usize) -> MatchBytes<'t> {
+        MatchBytes::new(self, start, end)
+    }
+    fn advance_position(&self, i: usize) -> usize {
+        next_input_pos(self, i)
```
keith-hall: is there any danger here that we don't check whether we are in ascii mode or not, and always use `codepoint_len`  inside `next_input_pos`?

lu-zero: good catch! Let me go over it once more and add few tests to make sure all the new code is covered.

keith-hall (approved):

keith-hall:
Thanks!

keith-hall:
Note to self: I need to add a changelog entry describing the breaking API change

keith-hall:
It turned out that quite a few changes were needed to properly respect ASCII mode, unicode was basically forced previously. I added some additional test cases for the bugs I found.

PR #249, opened 2026-05-09
seek: skip pre-filter for post-optimize standalone lookahead

stefanobaghino:
## Summary

`optimize::optimize_trailing_lookahead` rewrites a standalone top-level `(?=X)` to `Concat([Group(Empty), X])` so the body `X` becomes explicit content (and may go through the Wrap arm when `X` is easy). For the seek pre-filter that rewrite is unsound — the original is a zero-width assertion. Feeding the body's over-approximation to `regex-automata`'s META engine has been observed to mis-report leftmost position on shapes like `[char-class]+<…(?:nested)?…>` (PikeVM and the lazy DFA both agree on a strictly earlier start), causing `Insn::Seek` to jump past the actual leftmost match and the surrounding pattern to emit a false negative.

The fix detects the post-optimize `Concat([Group(Empty), X])` shape in `build_seek_pattern` and emits no seek pattern, so the VM falls through to the standard `SplitUnanchored` byte-by-byte preamble — which preserves the original lookahead semantics. See the diff in `src/seek.rs::build_seek_pattern` for the early-return and the comment that records the rationale.

## How this surfaced

[syntect](https://github.com/trishume/syntect) consumes fancy-regex via `--features default-fancy`. Enabling `seek(true)` for syntect's per-rule regexes broke three `make syntest-fancy` assertions on otherwise-passing baselines:

- C++ template `<>` mis-scoped as `keyword.operator.comparison.c` instead of `punctuation.definition.generic.begin.c++` (`testdata/Packages/C++/syntax_test_cpp.cpp:843`)
- Objective-C++ same pattern via the C++ inheritance chain (`testdata/Packages/Objective-C/syntax_test_objc++.mm:830`)
- Makefile `(?={{rule_lookahead}})` failing to push the `expect-rule` context, dropping the `meta.function entity.name.function` outer scopes (`testdata/Packages/Makefile/syntax_test_makefile.mak:732, 740`)

All three patterns reach fancy-regex via the post-optimize shape described above. Bisecting the seek over-approximation against PikeVM and the lazy DFA confirmed `regex-automata`'s META engine was the source of the leftmost-first divergence — the over-approximation itself is sound, but META picks a non-leftmost match for it. With this patch, all three syntest assertions pass again.

Two regression tests are added in `seek::tests`: one asserts that no seek pre-filter is emitted for hard standalone `(?=X)` (while the trailing-lookahead form `a(?=(?!c)b)`, which keeps a non-empty group, still gets one); the other replays the C++ generic-type-call lookahead end-to-end through `captures_from_pos` with both `seek(false)` and `seek(true)`, asserting both return the same leftmost position.

### DIFF
--- a/src/seek.rs
+++ b/src/seek.rs
@@ -107,6 +107,34 @@ pub(crate) fn build_seek_pattern<'a>(
     buf: &mut String,
     precedence: u8,
 ) {
+    // `optimize::optimize_trailing_lookahead` rewrites a standalone top-level
+    // `(?=X)` to `Concat([Group(Empty), X])` so that the body `X` becomes
+    // explicit content (and may go through the Wrap arm when `X` is easy).
+    //
+    // For the seek pre-filter that rewrite is unsound: the original pattern is
+    // a zero-width assertion, but the rewritten `()X` requires `X` to actually
+    // be matched.  Building a seek over-approximation from `X` then asks
+    // `regex-automata::meta::Regex` to find the leftmost position where `X`
+    // matches.  The META engine's leftmost-first search has been observed to
+    // pick non-leftmost matches for some patterns of the form
+    // `[char-class]+<…(?:nested)?…>` (verified by direct comparison against
+    // PikeVM and the lazy DFA, both of which agree on a strictly earlier
+    // start position).  When that happens, `Insn::Seek` jumps the VM past
+    // valid match positions and the surrounding pattern emits a false
+    // negative.
+    //
+    // Skipping the seek pre-filter for this specific tree shape preserves
+    // the original lookahead semantics: the parser falls through the standard
+    // `SplitUnanchored` byte-by-byte preamble, which never skips a position.
+    if let Expr::Concat(children) = info.expr {
+        if let Some(first) = children.first() {
+            if let Expr::Group(inner) = first {
+                if matches!(**inner, Expr::Empty) {
+                    return;
+                }
+            }
+        }
+    }
     build_seek_pattern_impl(info, group_info_map, depth, buf, precedence, false);
 }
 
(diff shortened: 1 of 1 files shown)


stefanobaghino:
Refreshed measurements on the `syncat` / `microsoft/TypeScript/checker.ts` workload from [my earlier comment on #246](https://github.com/fancy-regex/fancy-regex/pull/246#issuecomment-4406114811), now with `main` and this PR side-by-side. Same harness (`hyperfine --warmup 2 --runs 10`, AC, M-class P-core, syntect at [trishume/syntect#627](https://github.com/trishume/syntect/pull/627), 3.2 MB input).

| build                            | mean       | σ        | vs onig |
| -------------------------------- | ---------- | -------- | ------- |
| `regex-onig`                     | 3.343 s    | 0.007 s  | 1.00×   |
| `regex-fancy` (no seek)          | 14.220 s   | 0.088 s  | 4.25×   |
| `regex-fancy` `seek(true)` on `main` (unsafe — 3 syntest assertions fail) | 6.243 s | 0.050 s | 1.87× |
| **`regex-fancy` `seek(true)` on this PR (safe)** | **9.257 s** | **0.126 s** | **2.77×** |

Two takeaways:

* The fix correctly re-clears `make syntest-fancy` (the C++ / ObjC++ template `<>` and Makefile `(?={{rule_lookahead}})` cases discussed on #246), so there's no debate about merging it on correctness grounds.
* On this workload it does cost ~3.0 s wall-clock relative to `main` (1.87× → 2.77× onig) — TypeScript syntaxes are saturated with top-level `(?=keyword|identifier|…)` lookaheads that all hit the post-optimize `Concat([Group(Empty), X])` shape this PR disables seek for, and the byte-by-byte `SplitUnanchored` fall-through is materially slower for those bodies than the (when leftmost-correct) DFA pre-filter.

To answer [the release-mode question from #246](https://github.com/fancy-regex/fancy-regex/pull/246#issuecomment-4407713030): yes, all numbers above (and on #246) are release builds — `cargo build --release` end-to-end, including the `[patch.crates-io]`-resolved local `fancy-regex` (verified via `cargo build -v` showing `-C opt-level=3` for both crates). No LTO; debug symbols enabled for samply but those don't affect runtime.

Recoverable headroom looks meaningful — the 3 s gap is exactly the per-pattern over-approximation work this PR conservatively skips. A tighter trigger that only disables seek for sub-shapes where regex-automata's META engine actually mis-orders leftmost (rather than for every standalone `(?=X)`), or a different over-approximation that doesn't feed META the problematic shape (e.g. `X`'s first-character class in lieu of `X`'s body), would let the safety fix coexist with most of the seek win. Happy to look into either as a follow-up if it's of interest.

keith-hall:
Thanks for this. I'm not 100% sure I follow though, please bear with me.

Let's say we have a pattern like `(?=X)`, which gets optimized to `()X`.
When the seek pattern is generated, the capture group itself is skipped, and the contents are empty, so we end up with just `X`. This is in general fine to seek for because we only seek to the earliest start position and don't care how many characters matched.

I think we should investigate this more, with the exact pattern and exact input which isn't returning the expected result, to understand if the problem is really regex-automata not returning the correct left-most match, or something else... Perhaps it just so happens that there is some hard node inside which we don't approximate correctly when seeking, and all the patterns affected just happen to have trailing positive lookaheads?

stefanobaghino:
Took the homework. For each of the three failing `make syntest-fancy` cases I instrumented syntect's `regex.rs::search` to run `seek(true)` and `seek(false)` side-by-side and log divergences, captured the exact pattern + haystack span verbatim at the moment of failure, then fed each post-`optimize_trailing_lookahead` over-approximation through `regex_automata::meta::Regex`, `nfa::thompson::pikevm::PikeVM`, and `hybrid::regex::Regex` on the same span — all from inside `fancy-regex/src/seek.rs::tests` on `main` (no #249) using `pub(crate) build_seek_pattern` + `populate_group_info_map` + `optimize`.

`expected_leftmost` = position the *original* pattern returns through fancy-regex with `seek(false)` (= what the syntest passes with).

| case | META | PikeVM | lazy DFA | expected |
|---|---|---|---|---|
| C++ `generic-type` (`C++.sublime-syntax:846`) on `syntax_test_cpp.cpp:843` | **`Some(32)` ❌** | `Some(16)` ✓ | `None` (compile failed) | 16 |
| ObjC++ `generic-type` (== C++) on `syntax_test_objc++.mm:830` | **`Some(32)` ❌** | `Some(16)` ✓ | `None` (compile failed) | 16 |
| Makefile `rule_lookahead` on `syntax_test_makefile.mak:732` | **`Some(2)` ❌** | `Some(0)` ✓ | `Some(0)` ✓ | 0 |

PikeVM finds the strictly earlier correct leftmost in all three cases; META overshoots. Lazy DFA agrees with PikeVM where it compiles and bails out at compile time on the C++/ObjC++ approximations (almost certainly NFA-too-large given the nested `[^(){}&;*^%=<>-]` repetition).

So the bisection is unambiguous: it's a `regex-automata` META leftmost-first divergence on these specific over-approximation shapes, exactly as the PR description hypothesises. The conservative fix in this PR sidesteps it cleanly and is correct on the merits.

Two follow-up directions worth considering once this lands:

1. Route the seek pre-filter through PikeVM (or hybrid DFA when it compiles) instead of META for the post-optimize `Concat([Group(Empty), X])` shape — slower per-call than META but still much cheaper than the byte-by-byte fall-through this PR currently falls back to (which is the source of the [3.0 s syncat regression](https://github.com/fancy-regex/fancy-regex/pull/249#issuecomment-4413452539)).
2. File the META leftmost-first divergence upstream with the smallest of the three reproducers (Makefile is cleanest — pure ASCII, lazy DFA agrees with PikeVM so it's unambiguously META, doesn't depend on the unrelated `\p{L}` → `\p{l}` lower-casing in `Expr::CharClass::to_str` that I noticed in the C++ approximation).

Verbatim approximation strings + haystacks for all three cases plus the diagnostic harness are in [stefanobaghino/fancy-regex@diagnostic-pr249](https://github.com/stefanobaghino/fancy-regex) (`src/seek.rs::tests::diagnostic_*`), happy to PR if useful for the file.

keith-hall:
Thanks, that is very helpful. I haven't had time to dig deep yet, but perhaps it is worth checking the other [available regex engines in regex-automata](https://docs.rs/regex-automata/latest/regex_automata/index.html#available-regex-engines) individually to see which is returning the undesired result, and perhaps opening a discussion on the regex-automata repository to see if they would consider the inconsistency a bug...

also, perhaps disabling the `nfa` feature for example (https://github.com/fancy-regex/fancy-regex/blob/main/Cargo.toml#L28) to force regex-automata to use a different engine could help - it would be interesting to compare performance too...

You can probably tell that I am a little reluctant to hamstring performance on fancy-regex just because of some odd behavior in regex-automata... If we can fix it upstream it could help everyone, not just fancy-regex.

keith-hall:
Tomorrow I plan to make a PR to fancy-regex to skip adding capture groups to seek patterns. I thought it already did this, but it was only for hard nodes. Then I will try with the examples/tests on your branch and see how it goes. If it doesn't help I guess we should merge this for now to make the seek filter behave correctly.

keith-hall:
https://github.com/fancy-regex/fancy-regex/pull/250

stefanobaghino:
Following up on [rust-lang/regex#1345](https://github.com/rust-lang/regex/issues/1345): an agent opened the fix as [rust-lang/regex#1346](https://github.com/rust-lang/regex/pull/1346) (`fix/meta-leftmost-1345` on the same branch I'd been pulling). I evaluated it against the same setup as my earlier benchmarks here (`make syntest-fancy` + syncat / `microsoft/TypeScript/checker.ts`, `hyperfine --warmup 2 --runs 10`, AC, M-class P-core, syntect at trishume/syntect#627 with `.seek(true)` and a local `[patch.crates-io]`).

**Diagnostic on `fancy-regex/main`** with #1346 applied:

| case | META (pre-fix) | META (with fix) | PikeVM | DFA | expected |
|---|---|---|---|---|---|
| C++ `generic-type` | `Some(32)` ❌ | **`Some(16)` ✓** | `Some(16)` | `None` | 16 |
| ObjC++ (== C++) | `Some(32)` ❌ | **`Some(16)` ✓** | `Some(16)` | `None` | 16 |
| Makefile `rule_lookahead` | `Some(2)` ❌ | **`Some(0)` ✓** | `Some(0)` | `Some(0)` | 0 |

`make syntest-fancy` (full suite, `.seek(true)`): **empty diff** against the known-failures baseline on `main` with #1346. The 3 assertions in this PR's description (`syntax_test_cpp.cpp:843` cols 29-30 + 39-40, `syntax_test_makefile.mak:732` + `:740` col 0-1) pass, no regressions elsewhere.

**Syncat checker.ts:**

| build | mean | σ | vs onig | syntest |
|---|---|---|---|---|
| `regex-onig` | 3.343 s | 0.007 s | 1.00× | passes |
| `main` + seek | 6.243 s | 0.050 s | 1.87× | 3 fail |
| this PR (#249) + seek | 9.257 s | 0.126 s | 2.77× | passes |
| **`main` + #1346 + seek** | **6.227 s** | **0.077 s** | **1.86×** | **passes ✓** |

The upstream fix recovers the full 3.0 s gap this PR's conservative workaround introduces on `checker.ts` — its perf is statistically indistinguishable from `main` + seek alone, with full correctness.

Since `regex-automata` is independently versioned and released, the natural path once #1346 merges and ships a 0.4.x patch is to bump fancy-regex's `regex-automata` floor in `Cargo.toml`, which lets consumers pick the fix up transitively without any code change here.

Marking this superseded but leaving it open as a fallback for the case where upstream is delayed and a workaround is still wanted in the meantime. Happy to keep the conservative-disable approach updated if needed.

PR #250, opened 2026-05-11
improve seek pattern generation and add tests

keith-hall:
- don't keep capture groups in the output pattern
- test that optimized patterns (with trailing positive lookaheads) generate as expected

### DIFF
--- a/src/seek.rs
+++ b/src/seek.rs
@@ -485,14 +485,15 @@ mod tests {
     fn get_seek_pattern(re: &str) -> String {
         let mut tree = Expr::parse_tree(re).unwrap();
         let requires_capture_group_fixup = optimize(&mut tree);
-        
+
         let info = analyze(
             &tree,
             AnalyzeContext {
                 explicit_capture_group_0: requires_capture_group_fixup,
                 ..AnalyzeContext::default()
             },
-        ).unwrap();
+        )
+        .unwrap();
 
         let mut group_info_map = Map::new();
         populate_group_info_map(&mut group_info_map, &info);
@@ -506,7 +507,10 @@ mod tests {
         // Simple backref with no positional anchors: group body is inlined verbatim.
         // The `(?:...)` wrapping comes from Concat precedence handling when stripping groups.
         assert_eq!(get_seek_pattern(r"(abc)\1"), "(?:abc)(?:abc)");
-        assert_eq!(get_seek_pattern(r"(?i)(abc)\1"), "(?:(?i:a)(?i:b)(?i:c))(?i:(?i:a)(?i:b)(?i:c))");
+        assert_eq!(
+            get_seek_pattern(r"(?i)(abc)\1"),
+            "(?:(?i:a)(?i:b)(?i:c))(?i:(?i:a)(?i:b)(?i:c))"
+        );
     }
 
     #[test]
@@ -562,14 +566,20 @@ mod tests {
     fn seek_pattern_easy_expr_with_capture_group_strips_group_wrapper() {
         // Easy expressions with capture groups drop the capture group wrappings.
         assert_eq!(get_seek_pattern(r"(abc)"), "abc");
-        assert_eq!(get_seek_pattern(r"(abc(def))(?<named>ghi)"), "(?:abc(?:def))(?:ghi)");
+        assert_eq!(
+            get_seek_pattern(r"(abc(def))(?<named>ghi)"),
+            "(?:abc(?:def))(?:ghi)"
+        );
     }
 
     #[test]
     fn seek_pattern_optimized_easy_expr_strips_group_wrapper() {
         // Optimized patterns lose the capture group wrappings
         assert_eq!(get_seek_pattern(r"(?=abc)"), "(?:abc)");
-        assert_eq!(get_seek_pattern(r"(h)(e)(l)(l)(o)\s*(?=world)"), r"(?:hello\s*)(?:world)");
+        assert_eq!(
+            get_seek_pattern(r"(h)(e)(l)(l)(o)\s*(?=world)"),
+            r"(?:hello\s*)(?:world)"
+        );
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


stefanobaghino:
Took #250 for a spin against the [3-engine harness](https://github.com/fancy-regex/fancy-regex/pull/249#issuecomment-4415079363) and a fresh syncat / `microsoft/TypeScript/checker.ts` bench. Same setup as before (`hyperfine --warmup 2 --runs 10`, AC, M-class P-core, syntect at trishume/syntect#627 with `.seek(true)` and a local `[patch.crates-io]` to the relevant fancy-regex branch).

**Diagnostic** on `seek_improvements`: the over-approximation is now clean (`(?:body)` instead of `()(?:body)`, capture groups dropped per the new tests), but META still picks the same wrong leftmost for all three failing cases:

| case | META (pre-#250) | META (#250) | PikeVM | DFA | expected |
|---|---|---|---|---|---|
| C++ `generic-type` | `Some(32)` ❌ | `Some(32)` ❌ | `Some(16)` ✓ | `None` | 16 |
| ObjC++ (== C++) | `Some(32)` ❌ | `Some(32)` ❌ | `Some(16)` ✓ | `None` | 16 |
| Makefile `rule_lookahead` | `Some(2)` ❌ | `Some(2)` ❌ | `Some(0)` ✓ | `Some(0)` ✓ | 0 |

**`make syntest-fancy` with `.seek(true)`** on #250: the 3 assertions from the PR description (`syntax_test_cpp.cpp:843` cols 29-30 + 39-40, `syntax_test_makefile.mak:732` + `:740` col 0-1) still fail — same as `main`.

**Syncat checker.ts** (3.2 MB):

| build | mean | σ | vs onig | syntest |
|---|---|---|---|---|
| `regex-onig` | 3.343 s | 0.007 s | 1.00× | passes |
| `main` + seek | 6.243 s | 0.050 s | 1.87× | **3 fail** |
| **`#250` + seek** | **5.863 s** | **0.056 s** | **1.75×** | **3 fail** |
| `#249` + seek | 9.257 s | 0.126 s | 2.77× | passes |
| **`#249` + `#250` + seek** | **9.158 s** | **0.039 s** | **2.74×** | passes |

So #250 alone is a worthwhile structural improvement (cleaner approximation, ~6% faster than `main` on this workload — the `()` wasn't free) but doesn't address the leftmost-ordering symptom. Combining #249 + #250 gives essentially #249 alone (within noise), as expected: where seek is disabled by #249 the cleanup in #250 has nothing to act on, and TypeScript's grammar is dominated by exactly that shape.

The underlying META leftmost-first bug is now upstream at [rust-lang/regex#1345](https://github.com/rust-lang/regex/issues/1345) — 5-byte ASCII repro (pattern `[^()]*(?:\([^()]*\))?[^()]*:` on haystack `$(:):` returns `2..3` from META, `0..5` from PikeVM and the lazy DFA). Once that lands, removing #249's standalone-lookahead disable should recover the gap. Until then, #249 still seems necessary for syntect-with-seek correctness; happy to land #250 in parallel since it's a clean perf+cleanup win that doesn't conflict with #249.

stefanobaghino:
Bench update after the upstream fix for [rust-lang/regex#1345](https://github.com/rust-lang/regex/issues/1345) went up as [rust-lang/regex#1346](https://github.com/rust-lang/regex/pull/1346). Cherry-picked just the two seek.rs commits from this PR (`858fc5b` + `b5fbf75`) onto `fancy-regex/main` to compare apples-to-apples; same harness as before.

| build | mean | σ | vs onig | syntest |
|---|---|---|---|---|
| `regex-onig` | 3.343 s | 0.007 s | 1.00× | passes |
| `main` + seek | 6.243 s | 0.050 s | 1.87× | 3 fail |
| this PR + seek (no #1346) | 5.863 s | 0.056 s | 1.75× | 3 fail |
| `main` + #1346 + seek | 6.227 s | 0.077 s | 1.86× | passes |
| **this PR + #1346 + seek** | **6.287 s** | **0.068 s** | **1.88×** | **passes** |

The interesting bit: this PR's ~6 % gain over `main` (5.863 → 6.243) **disappears** once #1346 is in place. Best explanation I have is that the gain wasn't from "smaller approximation → less work," it was from "smaller approximation → smaller window for META to land in the wrong place → fewer VM re-tries after a seek-miss." With META picking the correct leftmost on the first try, there's no re-work for the stripping to compensate for, and the two builds collapse to statistical noise of each other.

So the perf justification for this PR rests on the bug being present. With #1346 landed, the value of the change is purely structural — the approximation is cleaner and easier to reason about, which is still worth something, but it's no longer carrying a perf number behind it. Leaving the merge decision to you on that basis.

One small side observation while running this: the `seek_improvements` branch as published carries 9 not-yet-merged commits between `upstream/main` and `858fc5b`. When I ran the full `make syntest-fancy` suite on the branch tip plus #1346, ~15 assertions across other syntaxes failed (ASP, C#, CSS, Go template, Groovy, HTML, Java JSP, several JavaScript, Lisp, Markdown, …) that pass on `main` + this PR's two commits alone. So at least one of the intermediate commits is regressing tests in combination with the upstream fix. Cherry-picking just `858fc5b` + `b5fbf75` onto a fresh branch off `main` reproduces clean (no new failures, empty diff against the baseline).

Worth a separate look at those intermediate commits before they merge, but unrelated to this PR's actual change.

keith-hall:
Thanks for the heads up about the regression @stefanobaghino. I see it occurred when I added ascii support to backreferences instead of always using unicode semantics. I will fix it when I can. Once done and regex-automata is released with the meta prefilter fix, I will make a new fancy-regex release. Hopefully I will get chance to review your syntect PRs soon too

keith-hall:
It turns out it is a bug in regex-automata's `earliest` implementation (especially with regard to newlines I think, at least affecting `\z`), so I will undo that optimization for now. (Likely tomorrow some time.)
https://docs.rs/regex-automata/latest/regex_automata/struct.Input.html#method.earliest

PR #255, opened 2026-06-08
Implement RegexSet

keith-hall:
Supercedes #215 (and honors the feedback from that PR, keeping a note about how it differs from the regex crate's API etc.)

`RegexSet` allows efficient matching of multiple regular expression patterns against the same input text. This is particularly useful for applications like syntax highlighting, where many patterns need to be matched against each line of text.

How it works:

The `RegexSet` uses a hybrid approach to achieve good performance:
A multi-pattern DFA is built for parallel evaluation, to provide very fast candidate position matching with linear time complexity.

At the earliest candidate position:
- any **hard patterns** (those with backreferences, lookaround, etc.) which could match there are evaluated individually using a backtracking VM in anchored mode. These may have exponential time complexity in pathological cases.
- any **easy patterns** (those without backreferences, lookaround, etc.) are also individually run through the underlying regex crate in anchored mode, to resolve capture groups etc. which may have been skipped in multi-DFA mode.

`find_input` returns only matches at the earliest start offset. This is deliberate:
the caller can inspect all matches at that position and decide what to do when multiple patterns match. This allows for most flexibility, without having to cater for various scenarios in the `RegexSet` itself. The `Input` struct makes it easy for the caller to advance position in case of an empty match winning, and the `RegexInput` struct sits on top of that to make it possible to specify rules like whether `\G` should match or not.

The matches are iterated through in priority order - leftmost first wins, and in case of ties, then the patterns are returned in the order in which they were added to the set. This allows to only compute what is necessary for each match.


Keats:
I've let Fable try to port giallo to fancy-regex and it found a few issues (not all related to RegexSet directly) + wrote some repros:

1. Panic

```
    #[test]
    fn test_panic() {
        let set = RegexSet::new([r"foo", r"\bbar\b"]).unwrap();
        let matches = set.find_input(RegexInput::new("fooé bar")).unwrap();
        let m = matches.unwrap().next().unwrap().unwrap();
        println!("OK: pattern {} matched {:?}", m.pattern(), m.as_str());
    }
```
outputs:
```
thread 'regexset::tests::test_panic' (1288631) panicked at src/regexset.rs:370:22:
overlapping DFA search is infallible: no cache capacity limit, no quit bytes, Anchored::Yes always supported: MatchError(Quit { byte: 195, offset: 3 })
```

2. Quantifiers

```
fn onig_is_match(pattern: &str, haystack: &str) -> bool {
    onig::Regex::new(pattern).unwrap().find(haystack).is_some()
}

fn fancy_is_match(pattern: &str, haystack: &str) -> bool {
    let mut builder = fancy_regex::RegexOptionsBuilder::new();
    builder.oniguruma_mode(true);
    builder
        .build(pattern.to_string())
        .unwrap()
        .is_match(haystack)
        .unwrap()
}

fn main() {
    let cases = [
        // (pattern, haystack)
        (r"a\s+{0,1}b", "ab"),
        (r"a\s+{0,1}b", "a b"),
        (r"a\s+{0,1}b", "a  b"),
        (r"a\s+{0,1}b", "a {0,1}b"),
        (r"a\s+{1,0}b", "ab"),
        (r"a\s+{1,0}b", "a b"),
        (r"a\s+{1,0}b", "a {1,0}b"),
        // the cpp-macro grammar pattern that surfaced this
        (r"\s+{1,0}/\*[!*]+(?:(?:\n|$)|(?=\s))", " /*! doc"),
    ];

    println!("{:42} {:12} {:>6} {:>6}", "pattern", "haystack", "onig", "fancy");
    let mut divergences = 0;
    for (pattern, haystack) in cases {
        let onig = onig_is_match(pattern, haystack);
        let fancy = fancy_is_match(pattern, haystack);
        let marker = if onig == fancy { "" } else { "  <-- DIVERGES" };
        if onig != fancy {
            divergences += 1;
        }
        let haystack = format!("{haystack:?}");
        println!("{pattern:42} {haystack:12} {onig:>6} {fancy:>6}{marker}");
    }

    println!("\n{divergences} divergence(s)");
    std::process::exit(if divergences == 0 { 0 } else { 1 });
}
```
Now that pattern is non-sense but it does exist in at least the cpp vscode grammar. It looks like onig might attempt to "fix" something like `{1,0}`

3. \G optimization

It looks like there is an optimization when \G is at the start of a regex but that can cause issues:

```
use std::sync::Arc;
use fancy_regex::{RegexInput, RegexOptionsBuilder, RegexSet};

fn check(pattern: &str) {
    let mut builder = RegexOptionsBuilder::new();
    builder.allow_input_assertion_overrides(true);
    let re = builder.build(pattern.to_string()).unwrap();
    let set = RegexSet::from_regexes([Arc::new(re)], Default::default()).unwrap();

    // \G suppressed: the lookbehind should still match via the \s alternative
    let input = RegexInput::new("ab 5;").continue_from_previous_match_end(false);
    let found = set
        .find_input(input)
        .unwrap()
        .and_then(|mut m| m.next())
        .map(|m| m.unwrap().as_str().to_string());
    println!("{pattern:14} -> {found:?} (expected Some(\"5\"))");
}

fn main() {
    check(r"(?<=\G|\s)5");
    check(r"(?<=\s|\G)5");
}
```

outputs:
```
(?<=\G|\s)5    -> None (expected Some("5"))
(?<=\s|\G)5    -> Some("5") (expected Some("5"))
```

src/regexset.rs
```diff
@@ -0,0 +1,697 @@
...
+pub struct RegexSet {
+    regexes: Vec<Arc<Regex>>,
+    earliest_match_finder: RaRegex,
+    overlapping_dfa: Arc<dfa::DFA>,
+    overlapping_cache_pool: Arc<Pool<dfa::Cache, DfaCachePoolFactory>>,
+}
+
+#[derive(Clone, Debug)]
+/// Configuration for a RegexSet
+pub struct RegexSetOptions {
```
Keats, line 160: Can we add public methods to set the sizes? Otherwise some regex set can't compile for giallo because they are too big. 

Also there are some internal regex crate caching causing some issues since we can't tweak it:

1. Some grammars don't compile at all without lifting the NFA limit (emacs-lisp)

```
RaConfig::new()
    .match_kind(MatchKind::LeftmostFirst)
    .nfa_size_limit(None)  // this allows actually building emacs-lisp
    .hybrid_cache_capacity(64 * (1 << 20));  // this makes it 1000x faster than the default
```

2. Some grammars don't compile because of the DFA cache check (wolfram):
```
  overlapping_dfa_builder
      .configure(
          dfa::Config::new()
              .match_kind(MatchKind::All)
              .unicode_word_boundary(compile_options.unicode)
              // needed to build
              .skip_cache_capacity_check(true),
      )
```
otherwise (not sure why it mentions lookbehind though):
```
  Failed to build reverse DFA for variable-length lookbehind:
  given cache capacity (2097152) is smaller than minimum required (4867652)
```

keith-hall:
Thanks for the feedback @Keats - I have fixed most of these issues now. I plan to work on public methods to set the sizes soon.

src/regexset.rs
keith-hall: the misleading error has now been fixed and some public api methods to control the sizes

Keats:
2 remaining things it seems:

1. stacked quantifiers

```rust
fn onig(p: &str, h: &str) -> Option<bool> {
    Some(onig::Regex::new(p).ok()?.find(h).is_some())
}

fn fancy(p: &str, h: &str) -> Option<bool> {
    let mut b = fancy_regex::RegexOptionsBuilder::new();
    b.oniguruma_mode(true);
    b.build(p.to_string()).ok()?.is_match(h).ok()
}

fn main() {
    let cases = [("x{3,0}x", "xx"), ("^x{2}+$", "xxxx"), ("x{0,3}+x", "xx")];
    let mut divergences = 0;
    for (p, h) in cases {
        let (o, f) = (onig(p, h), fancy(p, h));
        let mark = if o == f { "" } else { divergences += 1; "  <-- DIVERGES" };
        println!("{p:10} {h:6} onig={o:?} fancy={f:?}{mark}");
    }
    println!("\n{divergences} divergence(s)");
    std::process::exit(if divergences == 0 { 0 } else { 1 });
}

```

gives 
```
x{3,0}x    xx     onig=Some(false) fancy=Some(true)  <-- DIVERGES
^x{2}+$    xxxx   onig=Some(true) fancy=Some(false)  <-- DIVERGES
x{0,3}+x   xx     onig=Some(true) fancy=Some(false)  <-- DIVERGES
```


2. \G

```rust
/// onig leftmost search with `\G` suppressed (NOT_BEGIN_POSITION).
fn onig(p: &str, h: &str) -> Option<String> {
    let re = onig::Regex::new(p).ok()?;
    let mut region = onig::Region::new();
    let opts = onig::SearchOptions::SEARCH_OPTION_NOT_BEGIN_POSITION;
    re.search_with_options(h, 0, h.len(), opts, Some(&mut region))?;
    let (s, e) = region.pos(0)?;
    Some(h[s..e].to_string())
}

/// fancy anchored RegexSet search with `\G` suppressed.
fn fancy(p: &str, h: &str) -> Option<String> {
    let mut b = RegexOptionsBuilder::new();
    b.allow_input_assertion_overrides(true);
    let re = b.build(p.to_string()).ok()?;
    let set = RegexSet::from_regexes([Arc::new(re)], Default::default()).ok()?;
    let input = RegexInput::new(h).continue_from_previous_match_end(false);
    set.find_input(input)
        .ok()?
        .and_then(|mut m| m.next())
        .map(|m| m.unwrap().as_str().to_string())
}

fn main() {
    let cases = [(r"(?<!\G)b", "ab"), (r"(?<=\G|\s)5", "ab 5;")];
    let mut divergences = 0;
    for (p, h) in cases {
        let (o, f) = (onig(p, h), fancy(p, h));
        let mark = if o == f { "" } else { divergences += 1; "  <-- DIVERGES" };
        println!("{p:12} {h:7} onig={o:?} fancy={f:?}{mark}");
    }
    println!("\n{divergences} divergence(s)");
    std::process::exit(if divergences == 0 { 0 } else { 1 });
}
```

gives

```
(?<!\G)b     ab      onig=Some("b") fancy=None  <-- DIVERGES
(?<=\G|\s)5  ab 5;   onig=Some("5") fancy=Some("5")
```


The first issue doesn't seem to affect the tests in giallo but the second one does change the result for a couple of grammars.

keith-hall:
`\G` is fixed, I will look at quantifiers separately over the weekend

keith-hall:
I believe I have a fix for the quantifier divergences: https://github.com/fancy-regex/fancy-regex/pull/258

Issue #162, opened 2025-05-12
How similar to Oniguruma do we aim to be?


keith-hall:
Now that Oniguruma has been archived, there is a fair amount of interest in using fancy-regex as a replacement:
- https://github.com/trishume/syntect/issues/583
- https://github.com/sharkdp/bat/issues/3285

I feel like it could be a good time to discuss the project's goals. My personal aim is for fancy-regex to be able to be used in a syntax highlighting context where Oniguruma has been traditionally used, and to get correct results with decent performance.
The question is, do we want to take it a step further and have fancy-regex become a "drop-in replacement" (if that is even possible) for Oniguruma in other contexts as well (i.e. achieve test parity)?

For example, Oniguruma has some interesting behaviors, like the way the name of a capture group can be reused, which affects the way backreferences work as well as the match results available via the API.
Do we want to copy such functionality for compatibility or - if the primary motivation is for syntax highlighting - is it worth checking what is actually used in the wild (in various tmLanguage grammars etc) before deciding?

If we do decide to behave more like Oniguruma, I would like us to have more detailed documentation, to be more user friendly and perhaps set us apart a little. Maybe this example is kinda obvious but I had a look at inlining non-recursive subroutine calls, to get a performance boost by having the inlined pattern go straight through the regex crate if none of the expressions are hard / require backtracking, and I discovered that calling a subroutine also populates the capture group - which put a spanner in the works...

What do others think? 🙂

si14:
My two cents: there's a related discussion here https://github.com/getzola/zola/issues/2758

In particular, (oniguruma-to-es)[https://github.com/slevithan/oniguruma-to-es] can be interesting in two respects:
* as a guideline to which subset of Oniguruma features is actually used in real world tmLanguage grammars
* as a demonstration of "transpilation" of Oniguruma to ES regexes

One way forward would be to use ES as a goal and provide a `oniguruma-to-es`-esque compatibility layer. Note that `oniguruma-to-es` had to add a [runtime shim](https://github.com/slevithan/oniguruma-to-es/blob/main/src/subclass.js) on top of standard ES regexes to cover almost all tmLanguage they test on.

robinst:
> My personal aim is for fancy-regex to be able to be used in a syntax highlighting context where Oniguruma has been traditionally used, and to get correct results with decent performance.

That sounds like a good goal.

> do we want to take it a step further

[...]

PR #260, opened 2026-06-26
In Oniguruma parsing mode, reject StartLine matches at EOF following a trailing newline

keith-hall:
I think this should be used instead of `disallow_empty_match_at_eof_after_newline`, judging from https://github.com/fancy-regex/fancy-regex/issues/162#issuecomment-4804691364, although we might need to keep some of that behavior, not sure... CC @Keats

### DIFF
--- a/playground/src/lib.rs
+++ b/playground/src/lib.rs
@@ -313,8 +313,12 @@ fn info_to_tree_node<'a>(
                 EndTextIgnoreTrailingNewlines { crlf: false } => {
                     ("EndTextIgnoreTrailingNewlines", None)
                 }
-                StartLine { crlf: true, .. } => ("StartLine", Some("(crlf)".to_string())),
-                StartLine { crlf: false, .. } => ("StartLine", None),
+                StartLine { crlf: true } => ("StartLine", Some("(crlf)".to_string())),
+                StartLine { crlf: false } => ("StartLine", None),
+                StartLineOniguruma { crlf: true } => {
+                    ("StartLineOniguruma", Some("(crlf)".to_string()))
+                }
+                StartLineOniguruma { crlf: false } => ("StartLineOniguruma", None),
                 EndLine { crlf: true } => ("EndLine", Some("(crlf)".to_string())),
                 EndLine { crlf: false } => ("EndLine", None),
                 LeftWordBoundary => ("LeftWordBoundary", None),
--- a/playground/src/lib.rs
+++ b/playground/src/lib.rs
@@ -313,8 +313,8 @@ fn info_to_tree_node<'a>(
                 EndTextIgnoreTrailingNewlines { crlf: false } => {
                     ("EndTextIgnoreTrailingNewlines", None)
                 }
-                StartLine { crlf: true } => ("StartLine", Some("(crlf)".to_string())),
-                StartLine { crlf: false } => ("StartLine", None),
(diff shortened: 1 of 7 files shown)


Keats:
This breaks a few grammars, some examples regexes that don't compile anymore:

- `(?<=(?:\P{word}|^)of)(?!\p{word})`
- `(?<!(?:^|[^$._[:alnum:]])import)(?=;|$|^`
- `(?<=(^[\t ]*|[\t ]));`

Keats:
I let Claude work on this PR to make it the same as onig and it came up with the following patch:

```
     +++ src/analyze.rs 2026-06-26 10:31:34.159860636 +0200
     @@ -136,6 +136,15 @@
          /// When true, assertions with runtime input suppression overrides are promoted to hard
          /// so they run on the VM.
          allow_input_assertion_overrides: bool,
     +    /// True while analyzing inside a lookaround. Oniguruma's `^` (StartLine with
     +    /// `reject_after_trailing_newline_at_eof`) only rejects the empty match at the
     +    /// absolute end of the haystack when it anchors the match itself; used as a
     +    /// sub-assertion inside a lookaround it behaves as a plain line start (this is
     +    /// what real Oniguruma does — verified against the vscode-textmate fixtures).
     +    /// Treating it as the (delegatable) plain line start inside lookarounds also
     +    /// keeps variable-length lookbehinds that contain `^` compilable instead of
     +    /// forcing the whole body onto the backtracking VM.
     +    in_lookaround: bool,
      }

      impl<'a> Analyzer<'a> {
     @@ -155,6 +164,18 @@
              let mut const_size = false;
              let mut hard = false;
              match *expr {
     +            // Inside a lookaround, Oniguruma's `^` reverts to a plain (delegatable)
     +            // line start: the trailing-newline-at-EOF rejection only applies when `^`
     +            // anchors the match itself, not when it is a sub-assertion of a
     +            // lookahead/lookbehind (matches real Oniguruma, per the vscode-textmate
     +            // fixtures). This also keeps variable-length lookbehinds containing `^`
     +            // compilable instead of forcing the whole body onto the backtracking VM.
     +            Expr::Assertion(Assertion::StartLine {
     +                reject_after_trailing_newline_at_eof: true,
     +                ..
     +            }) if self.in_lookaround => {
     +                const_size = true;
     +            }
                  Expr::Assertion(assertion) if assertion.is_always_hard() => {
                      const_size = true;
                      hard = true;
     @@ -249,8 +270,11 @@
                  }
                  Expr::LookAround(ref child, _) => {
                      // NOTE: min_pos_in_group might seem weird for lookbehinds
     +                let was_in_lookaround = self.in_lookaround;
     +                self.in_lookaround = true;
                      let child_info =
                          self.visit(child, min_pos_in_group, inside_zero_rep, enclosing_group)?;
     +                self.in_lookaround = was_in_lookaround;
                      // min_size = 0
                      const_size = true;
                      hard = true;
     @@ -850,6 +874,7 @@
              find_not_empty,
              disallow_empty_match_at_eof_after_newline,
              allow_input_assertion_overrides,
     +        in_lookaround: false,
          };

          let mut analyzed = analyzer.visit(&tree.expr, 0, false, 0)?;
```

which does pass all tests. Basically just disable the rejection when it's inside a lookaround.

Keats:
LGTM!

PR #261, opened 2026-06-28
fix `\G` matching unexpectedly when used in a RegexSet

### LINKED ISSUE #162: How similar to Oniguruma do we aim to be?

keith-hall:
fixes https://github.com/fancy-regex/fancy-regex/issues/162#issuecomment-4826308777
CC @Keats

### DIFF
--- a/src/regexset.rs
+++ b/src/regexset.rs
@@ -924,4 +924,35 @@ mod tests {
         );
         assert_eq!(single, None);
     }
+
+    #[test]
+    fn continue_from_prev_match_works_as_expected_when_match_is_at_search_start() {
+        use crate::Arc;
+        use crate::RegexBuilder;
+
+        let (pat, hay) = (r"\Gx", "yx");
+
+        let re = RegexBuilder::new(pat).build().unwrap();
+        let single = re
+            .find_input(RegexInput::new(hay).from_pos(1))
+            .unwrap()
+            .map(|m| (m.start(), m.end()));
+
+        // Same regex, same input, via a RegexSet:
+        let set = RegexSet::from_regexes([Arc::new(re)], Default::default()).unwrap();
+        let via_set = set
+            .find_input(RegexInput::new(hay).from_pos(1))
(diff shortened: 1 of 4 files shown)


src/regexset.rs
```diff
@@ -131,7 +131,7 @@ use regex_automata::PatternID;
 use regex_automata::PatternSet;
 
 use crate::compile::options_to_rabuilder;
-use crate::vm::OPTION_ANCHORED;
+use crate::vm::{OPTION_ANCHORED, OPTION_SKIPPED_EMPTY_MATCH};
```
keith-hall: Probably it would make sense to rename this flag, as the naming doesn't quite suit this use case...

PR #264, opened 2026-07-04
Improve compilation performance

keith-hall:
ref https://github.com/fancy-regex/fancy-regex/issues/162#issuecomment-4843699037, big thanks to @Keats

see individual commits for details


Keats:
Note that I didn't look at the changes at all here, pure claude only.

keith-hall:
That's fine, thanks for the disclaimer. I have looked them over a little and they seem reasonable. I will leave the PR up for a short while to give others a chance to take a look and also to dig deeper myself when I get time. As it has benchmarks and Giallo proves it works across a wide variety of patterns, it feels like the changes are safe at least.

PR #265, opened 2026-07-05
Reduce catastrophic backtracking

keith-hall:
In the optimization pass, rewrite patterns of the form `Repeat(X) -> Repeat(lo=0, ...) -> Repeat(X)` into a leading `Repeat(X)`  followed by an optional tail where the middle repeat is made mandatory (lo raised from 0 to 1) and followed by the trailing `Repeat(X)`. This keeps semantics identical but removes the possibility of catastrophic backtracking.

In addition, it does a second rewrite for enclosing quantifiers (`+` and `*` specifically) around the transformed `Concat`, so patterns like:

- `(?:\s*\w?\s*)+` become `\s*(?:\w{1}\s*)*`
- `(?:\s*\w?\s*)*` become `(?:\s*(?:\w{1}\s*)*)?`

As a separate but semi-related optimization, unnecessarily repeated patterns also merge the quantifiers for the same goal. Example: `(?:x+)*` is identical to `x*`

Full disclaimer: I got Copilot to do most of the work, and iterated with followup prompts, and manually fixed a few things and then fixed up the original commits. The ideas and examples were mine, the implementation was done by the AI.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -64,6 +64,7 @@ See the @CONTRIBUTING.md guide for details.
 2. Maintain existing code structure and organization unless asked to reorganize
 3. Write unit tests for new functionality
 4. Document public APIs and complex logic. Suggest changes to the Markdown documents when appropriate
+5. If you can avoid unnecessary cloning by doing a mem swap, please do. Performance is a feature.
 
 ## Tests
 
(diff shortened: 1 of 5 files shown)


keith-hall:
I notice some of these optimisations break some tests in Giallo, I will investigate. Marking as a draft for now.

keith-hall:
All green again 👍

PR #277, opened 2026-09-05
Memoize delegated engine builds within a compile

keith-hall:
Building a regex-automata meta::Regex per delegated fragment dominates compile time for patterns with many easy fragments, and alternation branches (variable-lookbehind branches in particular) frequently produce the exact same fragment several times within one pattern. Profiling the VSCode TypeScript TextMate grammar showed 68% of engine-build time going to duplicates: one 6.5KB pattern built the same `=[^...]` fragment five times, and `\s*` was rebuilt ten times across the pattern set.

The compiler now keeps a per-compile map from the delegate pattern string (plus whether capture slots are compiled in) to the built engine, and identical fragments share it via a cheap Arc clone. The engine is fully determined by that key: options are fixed for a compile, and outer capture-group numbering lives in the Delegate's capture_groups range, not in the engine itself.


oech3:
Is it able to cut new stable release for this? uutils/coreutils want to use this.

keith-hall:
Done - 0.19.1 released

oech3:
Thankyou!

PR #278, opened 2026-09-05
fix: hex escapes above 0x7F match raw byte in bytes modes

### LINKED ISSUE #274: BytesMode::Ascii encodes high-byte \xNN literals as UTF-8
When `BytesMode::Ascii` is selected, a hexadecimal escape above `\x7F` should match the corresponding single input byte. Instead, `\xFF` is parsed as the Unicode scalar `U+00FF` and is subsequently lowered with UTF-8 encoding. The  compiled literal is therefore `C3 BF`, so it does not match a haystack containing the one byte `FF`.

Reproducer for 0.0.19 (the assertion fails):

```rust
use fancy_regex::{BytesMode, RegexBuilder};

let re = RegexBuilder::new(r"\xFF")
    .bytes_mode(BytesMode::Ascii)
    .build()
    .unwrap();

assert!(re.is_match(b"\xFF").unwrap());
```

_AI disclosure: LLMs have been used to detect the issue and write this description._

keith-hall:
`\xFF` (and fixed-width `\xHH` / brace `\x{HH}` with value which could differ in Ascii mode compared to Unicode) now emit a new `Expr::LiteralBytes` carrying the raw byte, instead of being forced through `char::from_u32` → UTF-8 encode. In Unicode mode the bytes are still re-encoded as Unicode scalars (preserving existing behaviour); in `BytesMode::Ascii` and `BytesMode::UnicodeBytes` they are matched verbatim.

Also added:
- `Insn::LitBytes` for raw-byte literals in the VM
- `emit_literal_bytes` helper in `Compiler`
- tests for Ascii, UnicodeBytes and Unicode hex-escape behaviour

These changes were made with the help of an LLM. Fixes #274

### DIFF
--- a/src/analyze.rs
+++ b/src/analyze.rs
@@ -90,6 +90,7 @@ impl<'a> Info<'a> {
     pub(crate) fn is_literal_get_casei(&self) -> Option<bool> {
         match *self.expr {
             Expr::Literal { casei, .. } => Some(casei),
+            Expr::LiteralBytes { .. } => None,
             Expr::Concat(_) => self.children.iter().try_fold(false, |any, child| {
                 child.is_literal_get_casei().map(|c| any || c)
             }),
@@ -217,6 +218,10 @@ impl<'a> Analyzer<'a> {
                 // right now each character in a literal gets its own node, that might change
                 min_size = 1;
                 const_size = literal_const_size(val, casei);
+            }
+            Expr::LiteralBytes { ref bytes, casei } => {
+                min_size = bytes.len();
+                const_size = literal_const_size("", casei);
             }
             Expr::Concat(ref v) => {
                 const_size = true;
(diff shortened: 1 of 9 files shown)


src/analyze.rs
```diff
@@ -218,6 +219,10 @@ impl<'a> Analyzer<'a> {
                 min_size = 1;
                 const_size = literal_const_size(val, casei);
             }
+            Expr::LiteralBytes { ref bytes, casei } => {
+                min_size = bytes.len();
+                const_size = literal_const_size("", casei);
```
keith-hall, line 224: I don't really like this - probably time to revisit the `literal_const_size` function...

keith-hall: so I looked into this more and regex-automata only handles simple Unicode case folding, whereas Oniguruma supports full Unicode case folding. Under simple case folding rules, it is always const size from what I understood.

PR #279, opened 2026-09-06
Fix capture_names() panic on zero-repeated named group

### LINKED ISSUE #275: `capture_names` panics on a pattern with a `{0}`-repeated named group
```rust
use fancy_regex::Regex;

fn main() {
    let re = Regex::new("(?<n>a){1}").unwrap();
    println!("{:?}", re.capture_names().collect::<Vec<_>>());

    let re = Regex::new("(a){0}").unwrap();
    println!("{:?}", re.capture_names().collect::<Vec<_>>());

    let re = Regex::new("(?<n>a){0}").unwrap();
    println!("{:?}", re.capture_names().collect::<Vec<_>>()); // panics
}
```

Output:

```
[None, Some("n")]
[None]

thread 'main' panicked at src/lib.rs:1678:18:
index out of bounds: the len is 1 but the index is 1
```

`Regex::new` accepts the pattern and it matches: `is_match("")` returns `Ok(true)`, and `captures("")` gives a group set of length 1 whose `name("n")` is `None`. Only `capture_names()` panics. `{0,0}` behaves the same way, and `(?<n>a){0}(?<m>b)` panics on `len is 2 but the index is 2`.

I would expect `capture_names()` to return an iterator for any pattern that compiled.

Tested on fancy-regex 0.19.0 and on current `main` (415eaab).

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

chiliec:
## What
Closes #275.

`capture_names()` panicked (`index out of bounds`) for any pattern with a named group inside a zero quantifier, e.g. `(?<n>a){0}`, `(?<n>a){0,0}`, or `(?<n>a){0}(?<m>b)`.

## Fix
A group inside `{0}` is dropped at compile time, so `captures_len()` shrinks but `named_groups` still holds the original (now out-of-range) index for that name. `capture_names()` indexed the `names` vector directly with that index and panicked. Guard the write with `get_mut`, so a stale index is skipped — matching how `Captures::name()` already tolerates it (it uses `.get()` and returns `None`).

## Tests
Added `capture_names_with_zero_repeated_named_group` in `tests/captures.rs` covering `(?<n>a){0}` and `(?<n>a){0,0}`.

## Validation
```
cargo test           # 291 + 46 + 51 + 80 + 57 + 39 + ... all green
cargo test --test captures capture_names   # 3 passed
cargo fmt --check    # clean
```
Verified genuine RED→GREEN: with the fix reverted the new test panics with `index out of bounds: the len is 1 but the index is 1`; with the fix it passes and the full suite stays green.

Happy to adjust.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ with the exception that 0.x versions can break between minor versions.
 ### Added
 ### Changed
 ### Fixed
+- Fix a panic in `capture_names()` for a pattern with a zero-repeated named group such as `(?<n>a){0}` (#275)
 
 ## [0.19.1] - 2026-09-06
 ### Added
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1675,7 +1675,9 @@ impl Regex {
         let mut names = Vec::new();
         names.resize(self.captures_len(), None);
         for (name, &i) in self.named_groups.iter() {
-            names[i] = Some(name.as_str());
+            if let Some(slot) = names.get_mut(i) {
+                *slot = Some(name.as_str());
+            }
         }
         CaptureNames(names.into_iter())
     }
--- a/tests/captures.rs
+++ b/tests/captures.rs
@@ -16,6 +16,17 @@ fn capture_names() {
 }
 
 #[test]
+fn capture_names_with_zero_repeated_named_group() {
+    let regex = common::regex("(?<n>a){0}");
+    let names: Vec<_> = regex.capture_names().collect();
+    assert_eq!(names, vec![None]);
+
+    let regex = common::regex("(?<n>a){0,0}");
+    let names: Vec<_> = regex.capture_names().collect();
+    assert_eq!(names, vec![None]);
+}
+
+#[test]
 fn captures_fancy() {
     let captures = common::assert_captures(r"\s*(\w+)(?=\.)", "foo bar.").unwrap();
     assert_eq!(captures.len(), 2);


DRMacIver:
This fixes the panic but I think the bug is bigger than this. Here's a further report from the Hegel agent I had working on this repo about a failure in this area that's still present in this branch:

Where a `{0}` group is followed by a named group, `capture_names()` still returns fewer entries than the match has groups and leaves out a name that `Captures::name` resolves:

```rust
use fancy_regex::Regex;

fn main() {
    for (pattern, haystack, name) in [
        ("(a){0}(?<g1>b)", "b", "g1"),
        ("(a){1}(?<g1>b)", "ab", "g1"),
    ] {
        let re = Regex::new(pattern).unwrap();
        let caps = re.captures(haystack).unwrap().unwrap();
        let groups: Vec<_> = (0..caps.len()).map(|i| caps.get(i).map(|m| m.range())).collect();
        println!("{pattern} on {haystack:?}");
        println!("  capture_names {:?}", re.capture_names().collect::<Vec<_>>());
        println!("  groups        {groups:?}");
        println!("  name({name:?})    {:?}", caps.name(name).map(|m| m.range()));
    }
}
```

Output:

```
(a){0}(?<g1>b) on "b"
  capture_names [None, None]
  groups        [Some(0..1), None, Some(0..1)]
  name("g1")    Some(0..1)
(a){1}(?<g1>b) on "ab"
  capture_names [None, None, Some("g1")]
  groups        [Some(0..2), Some(0..1), Some(1..2)]
  name("g1")    Some(1..2)
```

`(a){0}(?<g1>b)` has three groups and `name("g1")` gives the third of them, but `capture_names()` yields two entries and never mentions `g1`. With `{1}` in place of `{0}` there are three entries and `g1` sits at the index `name("g1")` points at. I expected `capture_names()` to yield one entry per group, so that entry `i` describes group `i`.

The count mismatch is already reachable on 0.19.1, on a pattern that does not panic there:

```rust
use fancy_regex::Regex;

fn main() {
    let re = Regex::new("(?<n>a){0}(b)").unwrap();
    let caps = re.captures("b").unwrap().unwrap();
    println!("capture_names {:?}", re.capture_names().collect::<Vec<_>>());
    println!("groups        {:?}", (0..caps.len()).map(|i| caps.get(i).map(|m| m.range())).collect::<Vec<_>>());
}
```

Output:

```
capture_names [None, Some("n")]
groups        [Some(0..1), None, Some(0..1)]
```

There are two entries for three groups, so zipping `capture_names()` against `Captures::iter()` misaligns.

DRMacIver:
BTW if you'd find them useful, here are the hegel tests https://gist.github.com/DRMacIver/73f6aef3845005044c2b08536fa678a6

(Claude output, I've not carefully verified)

chiliec:
Thanks @DRMacIver — you're right that the `{0}`-before-named-group case where `capture_names()` under-counts is a real, separate bug from the panic this PR fixes. This PR is deliberately scoped to stopping the panic (the crash was the immediate correctness/DoS issue), and I'd rather not widen it and risk holding up that fix.

I'm happy to take the `capture_names()` counting bug as a follow-up — I'll open a separate issue/PR referencing your report and the gist tests. Would you prefer I keep this one panic-only and land it first, then do the counting fix on top?

keith-hall:
Thanks for this. I'm wondering if it would be simpler to just update the compiler to stop dropping capture groups inside a zero-repetition (delegating that to regex-automata should be cheap, I imagine?), and potentially fix all related bugs in one go that way. Would it help with groups inside `(?(DEFINE)` though I wonder, or would that still be a problem?

chiliec:
Good question. I looked into the delegate-to-regex-automata idea and I think it's the right long-term fix but it's more involved than it first looks, so I'd still land this panic fix first.

The core problem is in how the `{0}` path allocates group numbers. Right now a zero-repeated group still consumes a group slot in `captures_len()` but its name gets attached to the wrong index (or dropped), so `capture_names()` is inconsistent with the actual captures. Concretely on this branch:

```
(?<a>x){0}(?<b>y)   captures_len=2  names=[None, Some("a")]      // should be 3 slots, "b" dropped
(a){0}(?<b>y)       captures_len=2  names=[None, None]           // "b" missing
(?<b>y)(?<a>x){0}   captures_len=2  names=[None, Some("b")]      // "a" dropped
```

So the panic is gone, but the counting bug you and @DRMacIver flagged is real and separate. Fixing it properly means reconciling group-number allocation through the zero-rep branch in `analyze.rs` with the `named_groups` map, not just guarding the index, which is why I'd rather not fold it into this PR.

On `(?(DEFINE)`: I think it stays a distinct case. DEFINE groups already get special handling (`Expr::DefineGroup`) since they define but never match, so they don't reach the engine the same way a normal `{0}` group does. A zero-rep delegation fix likely won't automatically cover DEFINE; that probably needs its own check.

Happy to take the counting fix as the follow-up I mentioned. I'll open a separate issue referencing DRMacIver's report and gist, with the cases above as tests. My preference is to land this panic-only fix first so the crash is closed, then do the counting fix on top. But if you'd rather I hold and do it all in one PR, I'll go that route, your call.

keith-hall (approved):
Thanks

keith-hall:
I raised a followup PR at https://github.com/fancy-regex/fancy-regex/pull/284 in case anyone is interested

PR #281, opened 2026-09-12
Add leftmost longest match mode

### LINKED ISSUE #280: Add leftmost longest match semantics
I suspect coreutils need leftmost-longest match semantics, so we should add support for it in the Regex Builders, behind an optional feature flag.
Ref:
- https://github.com/rust-lang/regex/discussions/1126#discussioncomment-7522624
- https://github.com/uutils/coreutils/issues/1145#issuecomment-1803504714
- https://github.com/uutils/grep/pull/117#pullrequestreview-5178171482

keith-hall:
Add an opt-in `leftmost_longest` mode that returns the longest match among all matches starting at the leftmost match position, matching POSIX leftmost-longest behavior. Closes #280

### DIFF
--- a/src/vm.rs
+++ b/src/vm.rs
@@ -114,8 +114,8 @@ pub(crate) const OPTION_NOT_CONTINUED_FROM_PREVIOUS_MATCH: u32 = 1 << 1;
 /// \K is ignored as part of this check - so empty matches can still be reported if the engine
 /// consumed characters and then \K was used afterwards.
 pub(crate) const OPTION_FIND_NOT_EMPTY: u32 = 1 << 2;
-/// When set, the VM uses leftmost-longest match semantics instead of leftmost-first.
 #[cfg(feature = "leftmost_longest")]
+/// When set, the VM uses leftmost-longest match semantics instead of leftmost-first.
 pub(crate) const OPTION_LEFTMOST_LONGEST: u32 = 1 << 3;
 
 // TODO: make configurable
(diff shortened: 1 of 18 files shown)


src/vm.rs
```diff
@@ -1058,6 +1083,18 @@ fn run_with<S: HaystackInput + ?Sized, T>(
                     if state.get(0) < match_range.start || state.get(1) > match_range.end {
                         break 'fail;
                     }
+                    #[cfg(feature = "leftmost_longest")]
+                    if leftmost_longest {
+                        let match_len = state.get(1) - state.get(0);
+                        if best_saves.is_none() || match_len > best_match_len {
+                            best_saves = Some(state.saves.clone());
```
lhecker, line 1090: FWIW it may be worthwhile to reuse `best_save`'s memory here. Unsafe code for the win: https://godbolt.org/z/xa7nha3jK

Edit: Actually, the number of saves is 2x the number of captures, right? And this path is very hot because it backtracks? In that case perhaps a small scalar loop here would cause a perf improvement.

lhecker (commented):
I did not notice that fancy-regex had its own VM. That means my assessment here was wrong and I apologize for that: https://github.com/uutils/coreutils/pull/14329#issuecomment-5634958486
Thank you for making this!

src/vm.rs
keith-hall: Thanks for the suggestions, I will merge this for now and make a release to unblock coreutils, then we can work on improving performance later.

PR #282, opened 2026-09-14
Giallo changes

Keats:
In the end I think it's nicer to have the byte set in fancy-regex since otherwise i still need to add regex-syntax to my crate and it needs to match the version from fancy-regex. It could be useful to other people as well I think?

I've also added 2 other things: 
- being able to disable delegate_prefilter on a regex like it is on regexset
- memoize parsed case insensitive fragments: some folding can be slow


keith-hall:
Thanks, makes sense. Question: does ByteSet only make sense when using RegexSet or is it useful without it? I'm considering putting RegexSet behind a Cargo feature flag, as some consumers of this library are binary-size sensitive and every little helps... Potentially we could use the same gate for ByteSet too?

Keats:
We could gate all non-essential stuff behind an `extra` feature, it doesn't bother me. I don't think byteset would be used exclusively for regexset though, I'm using it to _avoid_ building regexset.

keith-hall (approved):
Looks good to me, we can always gate it later. Perhaps worth adding something to the change log?

Keats:
CHANGELOG updated!

PR #283, opened 2026-09-19
Avoid inlining multiple times the same thing in seek

Keats:
The new benchmark compared to main:

```
compile_seek_recursion/fancy/seek_recursion_set
                        time:   [64.041 µs 64.088 µs 64.139 µs]
                        change: [-58.083% -58.021% -57.954%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
compile_seek_recursion/fancy/seek_recursion_seek_on
                        time:   [48.612 µs 48.655 µs 48.700 µs]
                        change: [-59.912% -59.876% -59.837%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Disclaimer: Claude found a bug with the initial patch and wrote the 2 tests


src/seek.rs
```diff
@@ -689,4 +696,32 @@ mod tests {
...
+        options.seek(true);
+        let actual = options
+            .build(pattern.to_string())
+            .unwrap()
+            .find(haystack)
+            .unwrap();
+
+        assert_eq!(
+            expected.map(|m| (m.start(), m.end())),
+            actual.map(|m| (m.start(), m.end()))
```
keith-hall, line 718: Just thinking, if we were to ever turn seek on by default, this test wouldn't catch bugs any more... Perhaps worth also explicitly checking the start of the match is position 0 as well as comparing equality between seek and no seek behavior?

keith-hall (commented):
Probably it would make sense to also include a benchmark for finding matches with seek enabled - such patterns like defined in `SEEK_RECURSION` may now compile faster, but perhaps matching those regex patterns would be slower on certain haystacks because the seek pattern can no longer rule out as many candidate start positions - it would be good to have it benchmarked so we can make an informed decision.

Keats:
There was a regression, it's been solved by emitting lazy quantifiers

keith-hall (approved):
Thanks

PR #287, opened 2026-09-21
Add RegexOptionsBuilder::required_bytes

Keats:
What's faster than running a regex?
Knowing whether it can match without even compiling it!


src/byte_set.rs
```diff
@@ -304,16 +311,79 @@ pub(crate) fn start_bytes(pattern: &str, options: &RegexOptions) -> Result<Optio
     Ok(byte_set_from_expr(&tree.expr, options))
 }
 
+fn required_byte_set_from_expr(expr: &Expr) -> ByteSet {
+    let mut byte_set = ByteSet::default();
+
+    match expr {
+        Expr::Literal { val, casei } => {
+            for byte in val.bytes() {
+                // Without, we can't build a byte set because eg `é` can show up as `É` and
```
keith-hall, line 320: I don't quite understand this comment, it feels like it is missing something after "without" - without what? 🙂

src/byte_set.rs
```diff
@@ -304,16 +311,79 @@ pub(crate) fn start_bytes(pattern: &str, options: &RegexOptions) -> Result<Optio
     Ok(byte_set_from_expr(&tree.expr, options))
 }
 
+fn required_byte_set_from_expr(expr: &Expr) -> ByteSet {
+    let mut byte_set = ByteSet::default();
+
+    match expr {
+        Expr::Literal { val, casei } => {
```
keith-hall, line 318: There is now also a `LiteralBytes` Expr added recently in https://github.com/fancy-regex/fancy-regex/pull/278, which will never be case insensitive - worth adding here?

src/byte_set.rs
```diff
@@ -304,16 +311,79 @@ pub(crate) fn start_bytes(pattern: &str, options: &RegexOptions) -> Result<Optio
...
+        _ => {}
+    }
+
+    byte_set
+}
+
+/// See [`crate::RegexOptionsBuilder::required_bytes`]
+pub(crate) fn required_bytes(pattern: &str, options: &RegexOptions) -> Result<ByteSet> {
+    let mut tree = Expr::parse_tree_with_flags(pattern, options.compute_flags())?;
+    optimize(&mut tree);
```
keith-hall, line 363: Is the optimization pass beneficial here?

keith-hall (commented):
This looks really nifty, nice one. I did a quick pass and have a few comments.

src/byte_set.rs
keith-hall: Thinking about it more, I wonder whether it is worth keeping the optimization pass, and returning the `ExprTree` in addition to the required bytes, so that if the bytes do appear in the haystack, we wouldn't need to pay the parsing cost twice when constructing the regex... (Would need more changes, I think to make a public API for compiling from a parse tree)

keith-hall: Related: #133

src/byte_set.rs
```diff
@@ -304,16 +311,79 @@ pub(crate) fn start_bytes(pattern: &str, options: &RegexOptions) -> Result<Optio
...
+        } => {
+            byte_set = required_byte_set_from_expr(true_branch);
+            let other = required_byte_set_from_expr(false_branch);
+            byte_set.intersection(&other);
+        }
+        Expr::Group(child) => return required_byte_set_from_expr(child),
+        Expr::AtomicGroup(child)
+        | Expr::LookAround(child, LookAround::LookAhead | LookAround::LookBehind)
+        | Expr::Repeat { child, lo: 1.., .. } => return required_byte_set_from_expr(child),
+        _ => {}
```
keith-hall, line 354: I worry that there is a danger here, that if/when/as we add more `Expr` types, we could forget to handle it here...

src/byte_set.rs
```diff
@@ -304,16 +311,79 @@ pub(crate) fn start_bytes(pattern: &str, options: &RegexOptions) -> Result<Optio
...
+                    byte_set.intersection(&other);
+                }
+            }
+        }
+        Expr::Conditional {
+            true_branch,
+            false_branch,
+            ..
+        } => {
+            byte_set = required_byte_set_from_expr(true_branch);
```
keith-hall, line 346: Technically the true branch needs the condition to match also

src/byte_set.rs
Keats: I don't think we need to optimize here.

> returning the ExprTree in addition to the required bytes, so that if the bytes do appear in the haystack, we wouldn't need to pay the parsing cost twice when constructing the regex... (Would need more changes, I think to make a public API for compiling from a parse tree)

I don't know, I don't think I would actually use that myself since I run those methods to get a byteset at a different time than compiling the regex.

src/byte_set.rs
Keats: For now I've put it as a no-op with the other enum kinds as I'm not super sure how to handle it

src/byte_set.rs
keith-hall: looking at it closer, I see we already have the `byte_set_from_expr` function which would be what could get promoted to public with #133, so it makes sense to not change anything else here.

Removing the optimization call is also a little dangerous in the sense that if someone decides to copy/paste from here later, they would miss it and might not intend to. If we do remove it, let's do so with an explicit comment about why we skip it.

src/byte_set.rs
keith-hall: I've pushed a commit with my idea, just need to add a test for it when I get time

Keats: That's what I had before but I wasn't sure it was that simple

keith-hall: Good point, you're right, it isn't always that simple - it depends which mode the compiler is in, which we don't know at parse time... Reverted.

keith-hall (approved):

